PR #37, opened 2019-06-22
Stub response passed to `prepareStream`’s callback to avoid a `TypeError`

smockle:
Attempting to fix https://github.com/dgreif/ring-alarm/issues/35#issuecomment-504610023

Disclaimer: I haven’t tested this locally, because I don’t have a Ring camera set up.

I inspected the call stack provided by @ifeign, and the cause of the `TypeError` appears to be [the attempt to read `response.address.type` on line 658](https://github.com/KhaosT/HAP-NodeJS/blob/a42102f5db0388b67f28974ca5fa489933ca1be6/lib/StreamController.js#L658) in HAP-NodeJS’ `StreamController`.

Because `ring-alarm-homebrige` [currently passes an `Error`](https://github.com/smockle/ring-alarm/blob/eb80eb3165b5412e689d078f2e417eb449bd6891/homebridge/camera-source.ts#L68) as `response`, and `Error` does not have a `address` property, attempting to read `type` from the non-existent `address` property results in `Cannot read property 'type' of undefined`.

### DIFF
--- a/homebridge/camera-source.ts
+++ b/homebridge/camera-source.ts
@@ -64,8 +64,42 @@ export class CameraSource {
 
   handleCloseConnection(connectionID: any) {}
 
-  prepareStream(request: any, callback: (response: any) => void) {
-    callback(new Error('Not implemented'))
+  prepareStream(
+    request: any,
+    callback: (response: {
+      address: {
+        type: 'v4' | 'v6'
+        ipAddress?: string
+      }
+      video: {
+        port?: number
+        ssrc?: number
+        proxy_pt?: string
+        proxy_server_address?: string
+        proxy_server_rtp?: string
+        proxy_server_rtcp?: string
+        srtp_key?: string
+        srtp_salt?: string
+      }
+      audio: {
+        port?: number
+        ssrc?: number
+        proxy_pt?: string
+        proxy_server_address?: string
+        proxy_server_rtp?: string
+        proxy_server_rtcp?: string
+        srtp_key?: string
+        srtp_salt?: string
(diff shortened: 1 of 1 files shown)


smockle:
Fixed issues that showed up when I ran `npm test`, amended the commit and force-pushed. When I have tests/lint rules passing, I’m planning to publish this as `homebridge-ring-alarm-smockle` for ease of testing.

smockle:
> I’m planning to publish this as `homebridge-ring-alarm-smockle` for ease of testing.
Done: https://www.npmjs.com/package/homebridge-ring-alarm-smockle

ifeign:
Just tested your branch @smockle, this seems to fix the error I was experiencing. Nothing appears in the logs when I select the camera and Homebridge does not crash. Thank you!

smockle:
@dgreif implemented this differently in https://github.com/dgreif/ring-alarm/commit/3fe39cd8106da178b8003bc55b50141c00e07fe7.

This PR passes a stubbed object to the callback function; that commit omits the call to the callback (replacing it with the comment: `// intentionally ignore the request.  This prevents "No Response" overlay for about 20s`). Both approaches should avoid the `TypeError`.

@dgreif feel free to close this PR. Thanks for your work on this! 😊

smockle:
> Just tested your branch @smockle, this seems to fix the error I was experiencing. Nothing appears in the logs when I select the camera and Homebridge does not crash. Thank you!

Also, @ifeign: thank you for taking the time to test this PR and for sharing the results. 😊

PR #104, opened 2019-08-12
Added the RingCamera.recordLiveVideoToFile function.

juanjose49:
Expanding on the work done by @bourdakos1, I created a function on the RingCamera class that allows users to record a live stream to file (video only, mp4). Unfortunately, usage is somewhat inconsistent for me because the RingCamera.createSipSession function seems to timeout half the time. Is there any chance we can add a timeout to this function or resolve the underlying issue causing the problem?

Thanks for your terrific work! I hope this PR serves the community well ;-)


dgreif:
Thanks for pulling this functionality over @juanjose49.  @bourdakos1 are you ok with me adding your code directly into the `ring` repo?  If so, were there any modifications you wanted to make before it gets merged in?

bourdakos1:
Yea I'm fine with it, the only reason I didn't originally add it was because I haven't 100% finished the packet decoder, but it should be good enough. There's also the option of just using ffmpeg, you can pipe the rtp stream directly and it will give you the mp4, or you can stream it to the browser as hls. The only caveat is that you need to have ffmpeg installed on your computer

bourdakos1:
It's also probably important to note that the h264 file that gets outputted by my code would need to get converted by ffmpeg or another tool before playing it

juanjose49:
@bourdakos, this pull request includes the addition of a npm dependency that installs ffmpeg and converts the h264 file to an mp4. 

Can you offer further explanation (or a link) on how to use ffmpeg directly to decode rtp packets?

bourdakos1:
Oh interesting, I didn't realize. Yea so then I recommend tossing my code haha. I can modify this to directly use the rtp stream

api/util.ts
```diff
@@ -77,3 +77,16 @@ export function stringify(data: any) {
...
 }
+
+/**
+ * Returns a Promise that will resolve after the number of seconds provided. Useful
+ * for keeping the async/await patterns in functions that require a simple sleep.
+ * 
+ * @param numberOfSeconds the number of seconds that this function will take to exit.
+ * @return {Promise<void>} a Promise that will resolve after the number of seconds provided.
+ */
+export function timeout(numberOfSeconds:number) : Promise<void>{
```
dgreif: This is duplicate of the `delay` method above, which takes milliseconds instead of seconds

examples/stream-to-file-example.ts
```diff
@@ -0,0 +1,35 @@
...
+      refreshToken: process.env.RING_REFRESH_TOKEN!
+    }),
+    [camera] = await ringApi.getCameras()
+
+  if (!camera) {
+    console.log('No cameras found')
+    return
+  }
+  // Get the absolute path to the examples directory.
+  const dirname = __dirname.split("/").slice(0,-2).join("/")+"/examples/recordings/";
```
dgreif: I would use `path.join` here instead since windows uses `\` for separator

api/h264-builder.ts
```diff
@@ -0,0 +1,115 @@
...
+        return
+      }
+      const nalType = (fragmentHeader & parseInt('00011111', 2)) >>> 0
+      if (start === 1) {
+        this.fileStream.write(Buffer.from([0x00, 0x00, 0x00, 0x01]))
+        const reconstructedHeader =
+          (forbidden << 7) | (nri << 5) | (nalType << 0)
+        this.fileStream.write(Buffer.from([reconstructedHeader]))
+      }
+      this.fileStream.write(message.subarray(payloadStartOffset + 2))
```
dgreif: @bourdakos1 is all of this code yours, or did you pull it from somewhere else?  I only ask because I'm curious and you knocked it out really fast after I released the first version of sip 😁

examples/stream-to-file-example.ts
juanjose49: People still use windows? Jk I'll fix it up.

api/h264-builder.ts
bourdakos1: Yea it's all mine, I was working on it for a while before you had the sip stuff done

bourdakos1:
I made a new pull request that decodes the rtp packets as H264 and pipes them over udp to ffmpeg. We could have ring send the rtp packets directly to ffmpeg, but we already have those ports set up, or we could proxy it, but this works well I think

examples/stream-to-file-example.ts
juanjose49: Addressed in subsequent commits.

dgreif:
@juanjose49 I really appreciate your efforts on this PR! I think @bourdakos1 did a really good job refactoring his code in #105, along with some more advanced stuff in the examples.  I'm going to close this PR in favor of #105

juanjose49:
Sounds good!

PR #105, opened 2019-08-13
ffmpeg save mp4

bourdakos1:


bourdakos1:
This adds 2 new examples:
- One that generates 10 second mp4 clips
- One that streams the footage to the browser

to test the browser example run `npm run browser-example` and open http://localhost:3000 in your browser

> **Note:** for both examples you need ffmpeg installed)

dgreif:
@bourdakos1 very nice cleanup and examples!  I haven't tried them out yet, but i'm excited to see tensorflow in the browser example.  Just a couple things before I merge:

* Could you remove the yarn lock file and update the npm lock?
* Do you think we should include the `PacketParser` in the main api, rather than just in the examples, since it is critical for h264 + ffmpeg streaming?

bourdakos1:
yea, sure thing

We should probably leave it out for now, I think we could sidestep using it in the future. Getting the raw rtp packets is more useful in my opinion

dgreif:
@bourdakos1 @juanjose49 this PR added some really good examples, but doesn't expose anything directly from the API for streaming to a file/browser.  I'm going to be working on piping the RTP directly to ffmpeg (skipping the PacketParser), and then setting up different configuration options for ffmpeg to output to the desired locations/streams.  I haven't made a ton of progress yet so let me know if either of you is eager to work on it.  I hope to have something solid in the next week or so, but I'm not sure how much free time I will have to work on it.  Thanks again for both of your efforts to bring these features directly into this library!

bourdakos1:
@dgreif I'd be happy to help out, but I'm not sure how much time I'll have this week to work on anything. I think the key to getting this to work is just saving the SDP information you pass in the SIP INVITE as a file and give that file to ffmpeg. I think the ffmpeg command is something like:
```
ffmpeg -protocol_whitelist "file,rtp,udp" -i <THE_SDP_FILE> <THE_REST_OF_THE_PARAMS_FOR_STREAMING_OR_CHUNKING_FILES>
```

dgreif:
@bourdakos1 that's the exact conclusion I came to after messing around with it last night! Thanks for confirming my thoughts.  I'll play around with it more over the next few days.

bourdakos1:
I originally just tried:
```
ffmpeg -i rtp://address:port out.mp4
```

but then I got errors because it didn't know packetization mode/ what do with packet type 99 etc... all the info is in the sdp stuff, and then I found you can just pass the sdp file and it will set up rtp and decode it.

but then I got errors about protocol not in whitelist and found this:
https://lists.ffmpeg.org/pipermail/ffmpeg-user/2016-February/030853.html

bourdakos1:
I knew you could do:
```
ffmpeg -i udp://address:port out.mp4
```
because I was working with tello drone streams and they just send the raw h264 over udp

bourdakos1:
I think when designing the api it would make sense to give the option for ffmpeg overrides, because there are a million tweaks that users could do and I definitely don't know enough about ffmpeg to give authoritative defaults. Especially when it comes to live-streaming theres a billion ways to do things and improve latency / quality etc...

maybe something like:
```
await camera.startLiveStream('output/stream.m3u8')

const override = ['-segment_time', '60'] // export 60 second parts instead of 10
await camera.startLiveVideoOutput('output/part%d.mp4', overrides)
```

juanjose49:
Giving users the ability to configure is great but introduces risks of
exploits. I'd keep the api simple and just count on people adding
additional explicit configurations via pull requests.

dgreif:
@juanjose49 I'm not sure what exploits users would introduce here...they are using the api with their own cameras on their own computer.  Can you give an example scenario where there could be an issue?  I was thinking the same way as @bourdakos1 that there are so many options I wouldn't want to lock users in..

juanjose49:
Maybe it’s far-fetched and depends on how you implement the configuration
passing. We’re running ffmpeg from a shell created in node and If you’re
naively passing configurations into ffmpeg a malicious actor could take
advantage of that to pass a “configuration” that is ‘&& create new user’.
They could ssh to the machine and it’s now compromised.



If a user is running the application on their machine with no online
interactions, what you’ve proposed is probably fine. But take a new
software engineer that builds a small webservice that exposes this
functionality as an API and is passing the configurations from the UI (for
some reason) then someone could pass the malicious configuration to the API
and gain access.



Configurability is always a great but when dealing with shells, it can be
risky. Just a thought.

bourdakos1:
@juanjose49 I think in this scenario it's pretty safe because we are using `spawn` so the args go directly to ffmpeg, not to the shell. I think we would run into the issue you mention if we used node's `exec` command, that just runs arbitrary code. Either way, I think this would most likely be serverside code so the frontend user shouldn't have access to submitting arbitrary commands to ffmpeg unless the developer specifically designed it that way. 
If for some reason a malicious user was able to get a malicious configuration injected, who's to say they couldn't get the api itself to run that code anyway

juanjose49:
Fair enough, sounds good!

Jafalex:
Hi there, thanks for an awesome plugin. I'm yet to try it with my freshly received ring doorbell. Just a quick question, how's #105 (and therefore #104) - which has been merged - any different from #107 which you understandably declined to dev? I'm just trying to understand what's possible with motion triggers and local storage... Also, would you consider just storing a static snapshot, or would the same reasoning as per #107 apply? Thanks again!

dgreif:
@Jafalex I think that's a very reasonable question.  The problem with #107 was that it would make it extremely easy for users to bypass the Ring subscription entirely, and just stream everything to their local storage.  #104 and #105 make that process _easier_, but it still requires someone to write code and set up the automation process themself if they wanted to bypass Ring.  If someone really wants to go to that much work, I'm not here to stop them.  What I don't want to do is make it so easy that users are _encouraged_ to use local storage and skip the Ring subscription.  Regarding snapshots, I'm still contemplating #130 which would allow you to save snapshots to Google Drive.  I think that provides some nice supplemental functionality without being a full replacement for the features you get from Ring.

Jafalex:
Thanks for your comprehensive answer @dgreif all makes sense. You have my vote for #130. Use case for us would be to give access to the drive folder to the Nanny so she can see what's up. She's not on iOS so isn't using Homekit and I don't want to give her access to the Ring account as she doesn't need to know what's going on when she's not looking after the kids... Cheers!

PR #140, opened 2019-09-28
Update config.schema.json

askovi:
Make password field input type password (hidden). This is for UI usage such as HOOBS and config-ui-x
Make email and password fields required while the rest remain optional.

### DIFF
--- a/homebridge/config.schema.json
+++ b/homebridge/config.schema.json
@@ -8,13 +8,18 @@
       "email": {
         "title": "Email",
         "type": "string",
+        "required": true,
         "format": "email",
         "placeholder": "Email"
       },
       "password": {
         "title": "Password",
         "type": "string",
-        "placeholder": "Password"
+        "placeholder": "Password",
+        "required": true,
+        "options": {
+            "hidden": true
+        }
       },
       "refreshToken": {
         "title": "Refresh Token",


homebridge/config.schema.json
```diff
@@ -8,13 +8,18 @@
       "email": {
         "title": "Email",
         "type": "string",
+        "required": true,
```
dgreif: Unfortunately we can't make email/password required.  The way I have everything set up, you can use _either_ `email` + `password` _or_ `refreshToken`.  The json schema form generator doesn't appear to have a mechanism for supporting this (if it does and I overlooked it, please let me know).  I have `oneOf` clause at the bottom that sound properly require one or the other, but it seems to be ignored.

homebridge/config.schema.json
```diff
@@ -8,13 +8,18 @@
...
         "placeholder": "Email"
       },
       "password": {
         "title": "Password",
         "type": "string",
-        "placeholder": "Password"
+        "placeholder": "Password",
+        "required": true,
+        "options": {
+            "hidden": true
```
dgreif: I'm fine with accepting this `options` section, however I'm not sure how useful it is since you can click over to the "Config" tab and see everything in plain text anyways

dgreif:
@askovi I think the only thing in here to keep is possibly the password field being marked as `hidden`.  Want to fix it up or just close this PR?

homebridge/config.schema.json
askovi: It is useful for HOOBS interface because we hide it in the UI unless you open the config file

dgreif:
Added in c0654495f2cf5da598625b23a63f5f2f7bae797a

PR #152, opened 2019-10-15
Enable or disable in-home doorbell (via API or Homebridge Switch)

smockle:
# Details
Toggles the “Ring my in-home doorbell” setting (screenshot below):
Ring my in-home doorbell

# Other Implementations
`python-ring-doorbell` implements this method in [`doorbot.py#L113-L133`](https://github.com/tchellomello/python-ring-doorbell/blob/6d2368b6615368a8724502d59755d14e0aa6e6f9/ring_doorbell/doorbot.py#L113-L133)

# Testing
I published [`homebridge-ring-smockle@5.8.8`](https://www.npmjs.com/package/homebridge-ring-smockle) to help with testing.

Note: If you toggle the switch in the Home app, then check status in the Ring app, you may see an outdated status, unless you’ve navigated to the main camera settings page.

# Example Use Cases
- Mute doorbell at night
- Mute doorbell when Do Not Disturb is enabled
- Mute doorbell when nursery white noise machine is turned on

### DIFF
--- a/api/ring-camera.ts
+++ b/api/ring-camera.ts
@@ -47,7 +47,7 @@ function getBatteryLevel(data: CameraData) {
   return batteryLevel
 }
 
-function getExistingDoorbellType({
+function getInHomeDoorbellType({
   settings: { chime_settings }
 }: CameraData): DoorbellType | undefined {
   if (!chime_settings || (!chime_settings.type && chime_settings.type !== 0)) {
@@ -69,7 +69,7 @@ export class RingCamera {
       this.batteryLevel !== null &&
       this.batteryLevel < 100 &&
       this.batteryLevel >= 0)
-  existingDoorbellType = getExistingDoorbellType(this.initialData)
+  inHomeDoorbellType = getInHomeDoorbellType(this.initialData)
 
   onRequestUpdate = new Subject()
   onRequestActiveDings = new Subject()
@@ -167,9 +167,9 @@ export class RingCamera {
     return true
   }
 
-  // Enable or disable the existing doorbell (if digital or mechanical)
-  async setExistingDoorbell(on: boolean) {
-    if (this.existingDoorbellType === undefined) {
+  // Enable or disable the in-home doorbell (if digital or mechanical)
+  async setInHomeDoorbell(on: boolean) {
+    if (this.inHomeDoorbellType === undefined) {
       return false
     }
 
--- a/api/ring-camera.ts
+++ b/api/ring-camera.ts
@@ -5,8 +5,7 @@ import {
   CameraHealth,
   HistoricalDingGlobal,
   RingCameraModel,
-  SnapshotTimestamp,
-  DoorbellType
+  SnapshotTimestamp
 } from './ring-types'
 import { clientApi, RingRestClient } from './rest-client'
 import { BehaviorSubject, Subject } from 'rxjs'
@@ -47,13 +46,11 @@ function getBatteryLevel(data: CameraData) {
   return batteryLevel
 }
 
-function getInHomeDoorbellType({
-  settings: { chime_settings }
-}: CameraData): DoorbellType | undefined {
-  if (!chime_settings || (!chime_settings.type && chime_settings.type !== 0)) {
+function getInHomeDoorbellStatus(data: CameraData): boolean | undefined {
+  if (!data.settings.chime_settings) {
     return undefined
   }
-  return chime_settings.type
+  return data.settings.chime_settings.enable
 }
 
 export class RingCamera {
@@ -69,7 +66,9 @@ export class RingCamera {
       this.batteryLevel !== null &&
       this.batteryLevel < 100 &&
       this.batteryLevel >= 0)
-  inHomeDoorbellType = getInHomeDoorbellType(this.initialData)
+  hasInHomeDoorbell =
+    this.initialData.settings.chime_settings &&
+    this.initialData.settings.chime_settings.type
 
   onRequestUpdate = new Subject()
   onRequestActiveDings = new Subject()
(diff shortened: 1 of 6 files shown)


api/ring-types.ts
```diff
@@ -295,26 +301,30 @@ export interface CameraData {
...
-      floodlight_settings?: any
-      light_schedule_settings?: any
-      luma_light_threshold?: number
     }
+    motion_snooze_preset_profile: string
+    live_view_preset_profile: string
+    live_view_presets: string[]
+    motion_snooze_presets: string[]
+    doorbell_volume: number
+    chime_settings: {
```
dgreif: Looks like this isn't defined on my stickup battery cam

```suggestion
    chime_settings?: {
```

api/ring-camera.ts
```diff
@@ -59,6 +73,7 @@ export class RingCamera {
       this.batteryLevel !== null &&
       this.batteryLevel < 100 &&
       this.batteryLevel >= 0)
+  existingDoorbellType = getExistingDoorbellType(this.initialData)
```
dgreif: It would probably be best to change this to an observable so that we can subscribe to it in the homebridge layer and to prevent stale data being accessible via the api

```suggestion
  onExistingDoorbellType = this.onData.pipe(
    map(getExistingDoorbellType),
    distinctUntilChanged()
  )
```

api/ring-camera.ts
```diff
@@ -46,6 +47,19 @@ function getBatteryLevel(data: CameraData) {
   return batteryLevel
 }
 
+function getExistingDoorbellType(data: CameraData): DoorbellType | undefined {
+  if (
+    !data ||
+    !data.settings ||
```
dgreif: From what I see in my devices, it looks like `data` and `data.settings` should always be defined.  Did you see any cases where they might not be?

api/ring-types.ts
```diff
@@ -295,26 +301,30 @@ export interface CameraData {
...
+    advanced_motion_detection_enabled: boolean
+    advanced_motion_detection_human_only_mode: boolean
+    luma_night_threshold: number
+    enable_audio_recording: boolean | null
+    people_detection_eligible: false
+    pir_settings?: any
+    pir_motion_zones?: number[]
+    floodlight_settings?: any
+    light_schedule_settings?: any
+    luma_light_threshold?: number
```
dgreif: Good catch with all these settings! Got em at the wrong level when I created the types I guess

api/ring-camera.ts
```diff
@@ -46,6 +47,19 @@ function getBatteryLevel(data: CameraData) {
   return batteryLevel
 }
 
+function getExistingDoorbellType(data: CameraData): DoorbellType | undefined {
```
dgreif: What do you think about using a different name? `Existing` is a bit vague, but I'm not sure what would be better.  `Physical`, `Mechanical`, `InHome`, `HardWired`...any other ideas?

api/ring-camera.ts
smockle: Nope—I have two cameras, and both have `data.settings`. I’ll remove the unnecessary checks, thanks!

smockle: Fixed in https://github.com/dgreif/ring/pull/152/commits/20324e41d9e21933e39a36c4bd152fdddb62fd8f

api/ring-types.ts
smockle: Thanks! Committed https://github.com/dgreif/ring/pull/152/commits/4789a682419e129cf5f382c18cb4e3ce84b2dee1 to fix.

api/ring-camera.ts
smockle: I agree! Switched the name to `InHome`, since it matches the text in the Ring app UI (“Ring my in-home doorbell”): https://github.com/dgreif/ring/pull/152/commits/8db456b598f1616431dd0c1158d5cfd641a9fe05

api/ring-camera.ts
smockle: Makes sense! I implemented this incorrectly in https://github.com/dgreif/ring/pull/152/commits/a7fa4a2eeade1c2d0ef4454a0927b74371d126fd — after reading through the `lock.ts` code I see what I ought to have done. Working to update this PR now.

smockle: Done—fixed in https://github.com/dgreif/ring/pull/152/commits/e13c4bb1a1ecff5f9cb71f539ca90348c241764c with a minor additional fix in https://github.com/dgreif/ring/pull/152/commits/08ce82871015b3df669703572bda9f253d33fbd1

api/ring-camera.ts
```diff
@@ -156,6 +171,35 @@ export class RingCamera {
...
+    this.updateData({
+      ...this.data,
+      settings: {
+        ...this.data.settings,
+        chime_settings: {
+          ...this.data.settings.chime_settings,
+          enable: on
+        }
+      }
+    })
```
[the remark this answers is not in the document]
smockle: Thanks! Fixed in https://github.com/dgreif/ring/pull/152/commits/a7fa4a2eeade1c2d0ef4454a0927b74371d126fd

smockle:
Thanks for the suggestions @dgreif; I appreciate you taking the time!

I’ve addressed the feedback from your initial review. I updated this PR and published a new version of the test package: [`homebridge-ring-smockle@5.8.8`](https://www.npmjs.com/package/homebridge-ring-smockle).

With the test package, when I toggle the “Ring my in-home doorbell” setting in the Ring app, I see the updated switch status reflected in the Home app, and vice versa.

dgreif:
Released in v`5.9.0`.  Thanks for the hard work on this, it worked great when I tested it with my doorbell.  Opens a world of possibilities for all those parents who keep putting "Don't Ring the Doorbell!" signs on their doors 😂

PR #269, opened 2020-03-10
Update for npm and npm download badges.

donavanbecker:
Changed the badges from using https://shields.io to using https://badgen.net which is a quicker generator


donavanbecker:
Also Added Homebridge Verified .svg file

donavanbecker:
@dgreif looks like you might have to change the readme a little to get the .svg to display.

PR #273, opened 2020-03-26
Update homebridge-ring README.md

mxdanger:
Fixed image not properly displaying.


apexad:
See https://github.com/dgreif/ring/pull/274 instead.  No need to use HTML.

PR #274, opened 2020-03-28
docs: fix branding svg & add homebridge:verified badge

apexad:
Small fixes /homebridge/README.md file
1. Correct `branding` svg so it shows up (correct way vs. https://github.com/dgreif/ring/pull/273 that uses pure markdown)
2. Add [verified-by-homebridge](https://github.com/homebridge/homebridge/wiki/Verified-Plugins) badge


apexad:
Closing, noticed that width cannot be controlled and still use markdown.

PR #331, opened 2020-05-22
Add support for thermostats

smockle:
# Context

I added a [Honeywell T6 Pro Z-Wave Programmable Thermostat](https://www.honeywellhome.com/us/en/products/air/thermostats/wifi-thermostats/t6-pro-z-wave-thermostat-th6320zw2003-u/) to my Ring Alarm Z-Wave network. After inclusion, a thermostat device, a temperature sensor and a “thermostat operating status” device appeared in the Ring app (screenshot of the thermostat below).

When `homebridge-ring` is restarted, the following message appears in the log:
```
[Ring] Hidden accessory temperature-control.thermostat Foyer Thermostat
```

In this PR, I attempt to add support for `temperature-control.thermostat` as a [Thermostat](https://developers.homebridge.io/#/service/Thermostat) platform accessory.

**Screenshot of thermostat controls in the Ring app:**
thermostat-in-ring-app

# Current Status

* The code lints and compiles, and is complete as far as I know
* My thermostat appears in the Home app (along with correct current temperature, set point, mode, and status)
* When I change set point or mode in the Home app, the physical unit reflects the changes
* When I change set point or mode using the physical unit, the Home app reflects the changes

# Previous Status

* ~The code compiles and lints, and everything I know to add is there~
* ~I haven’t yet seen the thermostat appear in HomeKit~
* ~Related: I haven’t yet controlled the thermostat via the Home app~
* ~I haven’t verified updates are handled properly when the physical controls on the thermostat are used~
* During my tests, I accidentally wiped all devices added via `homebridge-ring`—UUID’s have a [special prefix](https://github.com/dgreif/ring/blob/d1cf0a5d3eae4d506d7aea839b4bb73140a042a0/homebridge/ring-platform.ts#L227) when debugging, so non-prefixed devices are treated as [stale and removed](https://github.com/dgreif/ring/blob/d1cf0a5d3eae4d506d7aea839b4bb73140a042a0/homebridge/ring-platform.ts#L292-L300). ~I assume I can avoid this in the future by either removing the debug prefix in code, or backing up my `accessories` and `persist` directories.~ Update: @dgreif suggested [a better way to test](https://github.com/dgreif/ring/pull/331#issuecomment-632446143) (i.e. `npm run test-homebridge`)

### DIFF
--- a/api/location.ts
+++ b/api/location.ts
@@ -77,28 +77,8 @@ export class Location {
       }
 
       return deviceList.reduce((updatedDevices: RingDevice[], data) => {
-        const flatData = flattenDeviceData(data)
-
-        // Embed component device data in parent device data
-        // Used to read ambient temperature from the temperature sensor inside a thermostat
-        if (flatData.componentDevices) {
-          flatData.componentDevices = flatData.componentDevices.map(
-            ({ rel, zid }) => {
-              const componentDeviceData: RingDeviceData | undefined =
-                zid && deviceList.find((d) => flattenDeviceData(d).zid === zid)
-              return {
-                rel,
-                zid,
-                ...(componentDeviceData &&
-                  flattenDeviceData(componentDeviceData)),
-              }
-            }
-          )
-        }
-
-        const existingDevice = updatedDevices.find(
-          (x) => x.zid === flatData.zid
-        )
+        const flatData = flattenDeviceData(data),
+          existingDevice = updatedDevices.find((x) => x.zid === flatData.zid)
 
         if (existingDevice) {
           existingDevice.updateData(flatData)
--- a/api/location.ts
+++ b/api/location.ts
@@ -77,8 +77,28 @@ export class Location {
       }
 
       return deviceList.reduce((updatedDevices: RingDevice[], data) => {
-        const flatData = flattenDeviceData(data),
-          existingDevice = updatedDevices.find((x) => x.zid === flatData.zid)
+        const flatData = flattenDeviceData(data)
+
+        // Embed component device data in parent device data
+        // Used to read ambient temperature from the temperature sensor inside a thermostat
+        if (flatData.componentDevices) {
+          flatData.componentDevices = flatData.componentDevices.map(
+            ({ rel, zid }) => {
+              const componentDeviceData: RingDeviceData | undefined =
+                zid && deviceList.find((d) => flattenDeviceData(d).zid === zid)
+              return {
+                rel,
+                zid,
+                ...(componentDeviceData &&
+                  flattenDeviceData(componentDeviceData)),
+              }
+            }
+          )
+        }
+
+        const existingDevice = updatedDevices.find(
+          (x) => x.zid === flatData.zid
+        )
 
         if (existingDevice) {
           existingDevice.updateData(flatData)
--- a/api/ring-device.ts
+++ b/api/ring-device.ts
@@ -32,6 +32,11 @@ export class RingDevice {
     return this.data.name
   }
 
+  async getComponentDevices() {
+    const devices = await this.location.getDevices()
+    return devices.filter(({ data }) => data.parentZid === this.id)
+  }
+
   get supportsVolume() {
     return (
       deviceTypesWithVolume.includes(this.data.deviceType) &&
--- a/api/ring-device.ts
+++ b/api/ring-device.ts
@@ -1,6 +1,6 @@
 import { BehaviorSubject } from 'rxjs'
 import { deviceTypesWithVolume, RingDeviceData } from './ring-types'
-import { filter } from 'rxjs/operators'
+import { filter, reduce } from 'rxjs/operators'
 import { Location } from './location'
 
 export class RingDevice {
@@ -32,9 +32,15 @@ export class RingDevice {
     return this.data.name
   }
 
-  async getComponentDevices() {
-    const devices = await this.location.getDevices()
-    return devices.filter(({ data }) => data.parentZid === this.id)
+  get onComponentDevices() {
+    return this.location.onDevices.pipe(
+      reduce((acc, value) => {
+        const componentDevices = value.find(
+          (device) => device.data.parentZid === this.id
+        )
+        return acc.concat(componentDevices || [])
+      }, [] as RingDevice[])
+    )
   }
 
   get supportsVolume() {
--- a/api/ring-device.ts
+++ b/api/ring-device.ts
@@ -1,6 +1,6 @@
 import { BehaviorSubject } from 'rxjs'
 import { deviceTypesWithVolume, RingDeviceData } from './ring-types'
-import { filter, reduce } from 'rxjs/operators'
+import { filter, map } from 'rxjs/operators'
 import { Location } from './location'
 
 export class RingDevice {
@@ -32,13 +32,10 @@ export class RingDevice {
     return this.data.name
   }
 
-  getComponentDevices(callback: (componentDevices: RingDevice[]) => void) {
-    return this.location.onDevices.subscribe((devices) => {
-      const componentDevices = devices.filter(
-        ({ data }) => data.parentZid === this.id
-      )
-      callback(componentDevices)
-    })
+  get onComponentDevices() {
+    return this.location.onDevices.pipe(
+      map((devices) => devices.filter(({ data }) => data.parentZid === this.id))
+    )
   }
 
   get supportsVolume() {
--- a/api/ring-device.ts
+++ b/api/ring-device.ts
@@ -32,15 +32,13 @@ export class RingDevice {
     return this.data.name
   }
 
-  get onComponentDevices() {
-    return this.location.onDevices.pipe(
-      reduce((acc, value) => {
-        const componentDevices = value.find(
-          (device) => device.data.parentZid === this.id
-        )
-        return acc.concat(componentDevices || [])
-      }, [] as RingDevice[])
-    )
+  getComponentDevices(callback: (componentDevices: RingDevice[]) => void) {
+    return this.location.onDevices.subscribe((devices) => {
+      const componentDevices = devices.filter(
+        ({ data }) => data.parentZid === this.id
+      )
+      callback(componentDevices)
+    })
   }
 
   get supportsVolume() {
--- a/api/ring-device.ts
+++ b/api/ring-device.ts
@@ -9,6 +9,9 @@ export class RingDevice {
   id = this.zid
   deviceType = this.initialData.deviceType
   categoryId = this.initialData.categoryId
+  onComponentDevices = this.location.onDevices.pipe(
+    map((devices) => devices.filter(({ data }) => data.parentZid === this.id))
+  )
 
   constructor(
     private initialData: RingDeviceData,
@@ -30,12 +33,6 @@ export class RingDevice {
 
   get name() {
     return this.data.name
-  }
-
-  get onComponentDevices() {
-    return this.location.onDevices.pipe(
-      map((devices) => devices.filter(({ data }) => data.parentZid === this.id))
-    )
   }
 
   get supportsVolume() {
--- a/api/ring-types.ts
+++ b/api/ring-types.ts
@@ -112,7 +112,7 @@ export const RingCameraModel: { readonly [P in RingCameraKind]: string } = {
 }
 
 export type AlarmMode = 'all' | 'some' | 'none'
-export type ThermostatMode = 'heat' | 'cool' | 'off'
+export type ThermostatMode = 'heat' | 'cool' | 'off' | 'aux'
 export type MessageType =
   | 'RoomGetList'
   | 'SessionInfo'
--- a/api/ring-types.ts
+++ b/api/ring-types.ts
@@ -27,6 +27,7 @@ export enum RingDeviceType {
   BeamsTransformerSwitch = 'switch.transformer.beams',
   RetrofitBridge = 'bridge.flatline',
   RetrofitZone = 'sensor.zone',
+  Thermostat = 'temperature-control.thermostat',
 }
 
 export enum RingDeviceCategory {
@@ -216,6 +217,21 @@ export interface RingDeviceData {
   celsius?: number // no F provided, just celsius
   faultHigh?: number
   faultLow?: number
+  // temperature-control.thermostat
+  // Related: 'mode?: ThermostatMode' (above)
+  setPoint?: number
+  setPointMax?: number
+  setPointMin?: number
+  componentDevices?: (
+    | {
+        rel: string
+        zid: string
+      }
+    | ({
+        rel: string
+        zid: string
+      } & Omit<RingDeviceData, 'componentDevices'>)
+  )[]
 }
 
 export const deviceTypesWithVolume = [
--- a/api/ring-types.ts
+++ b/api/ring-types.ts
@@ -222,16 +222,10 @@ export interface RingDeviceData {
   setPoint?: number
   setPointMax?: number
   setPointMin?: number
-  componentDevices?: (
-    | {
-        rel: string
-        zid: string
-      }
-    | ({
-        rel: string
-        zid: string
-      } & Omit<RingDeviceData, 'componentDevices'>)
-  )[]
+  componentDevices?: {
+    rel: string
+    zid: string
+  }[]
 }
 
 export const deviceTypesWithVolume = [
--- a/api/ring-types.ts
+++ b/api/ring-types.ts
@@ -111,6 +111,7 @@ export const RingCameraModel: { readonly [P in RingCameraKind]: string } = {
 }
 
 export type AlarmMode = 'all' | 'some' | 'none'
+export type ThermostatMode = 'heat' | 'cool' | 'off'
 export type MessageType =
   | 'RoomGetList'
   | 'SessionInfo'
@@ -179,7 +180,7 @@ export interface RingDeviceData {
   locked?: 'jammed' | 'locked' | 'unlocked' | 'unknown'
   roomId?: number
   volume?: number
-  mode?: AlarmMode
+  mode?: AlarmMode | ThermostatMode
   transitionDelayEndTimestamp?: number | null
   alarmInfo?: {
     state: AlarmState
--- a/homebridge/thermostat.ts
+++ b/homebridge/thermostat.ts
@@ -124,7 +124,7 @@ export class Thermostat extends BaseDeviceAccessory {
         }
         // Documentation: https://developers.homebridge.io/#/characteristic/CurrentTemperature
         // 'Characteristic.CurrentTemperature' supports 0.1 increments
-        return Number(Number(temperature).toFixed(1))
+        return Math.round(temperature * 10) / 10
       },
     })
 
--- a/homebridge/thermostat.ts
+++ b/homebridge/thermostat.ts
@@ -9,8 +9,10 @@ import { RingPlatformConfig } from './config'
 import { hap } from './hap'
 
 export class Thermostat extends BaseDeviceAccessory {
-  private temperatureSensor: Observable<RingDevice>
-  private temperature: Observable<number | undefined>
+  private temperatureSensor: RingDevice | undefined
+  private temperature: Observable<
+    number | undefined
+  > = new Observable(({ next }) => next(0))
 
   constructor(
     public readonly device: RingDevice,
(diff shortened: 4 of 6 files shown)


homebridge/thermostat.ts
```diff
@@ -0,0 +1,126 @@
...
+        }
+      },
+      setValue: (mode: ThermostatMode) => {
+        this.logger.info(`Setting ${this.device.name} mode to ${mode}`)
+
+        return this.device.setInfo({ device: { v1: { mode } } })
+      },
+    })
+
+    this.registerCharacteristic({
```
smockle, line 57: @dgreif Do you have any guidance about when to use `registerCharacteristic` as opposed to `registerObservableCharacteristic`? I noticed `TemperatureSensor` [uses `registerObservableCharacteristic`](https://github.com/dgreif/ring/blob/d1cf0a5d3eae4d506d7aea839b4bb73140a042a0/homebridge/temperature-sensor.ts#L22) for the `CurrentTemperature` characteristic on the `TempSensorService` service.

dgreif:
I’ll follow up in the morning with more details, but for now I wanted to point out the npm run test-homebridge command. You can create a . homebridge directory in your project folder, and place a config.json in there with a different ID floor this test homebridge instance. That way you can add and remove the test bridge and it won’t interfere with your main instance. 

If the device is showing as hidden, make sure you add a new device type to the switch statement at the top of the platform file, which tells the platform how to handle devices of that specific type

smockle:
> I’ll follow up in the morning with more details, but for now I wanted to point out the npm run test-homebridge command. You can create a .homebridge directory in your project folder, and place a config.json in there with a different ID for this test homebridge instance. That way you can add and remove the test bridge and it won’t interfere with your main instance.

That’s super! I’ll use that for testing from now on. Thanks!

> If the device is showing as hidden, make sure you add a new device type to the switch statement at the top of the platform file, which tells the platform how to handle devices of that specific type

Thanks! Just added this in https://github.com/dgreif/ring/pull/331/commits/1a9115f9e4c533ab0617c23a9171eca07bbde995

api/location.ts
```diff
@@ -77,8 +77,28 @@ export class Location {
       }
 
       return deviceList.reduce((updatedDevices: RingDevice[], data) => {
-        const flatData = flattenDeviceData(data),
-          existingDevice = updatedDevices.find((x) => x.zid === flatData.zid)
+        const flatData = flattenDeviceData(data)
+
+        // Embed component device data in parent device data
```
smockle, line 82: This exists specifically to add additional `RingDeviceData` (i.e. is separate from any kind of HomeKit accessory linking). But it reminded me of Homebridge’s [`addLinkedService`](https://github.com/homebridge/HAP-NodeJS/blob/f3f032547f9c21f61c6338fbaf948b8258415012/src/lib/Service.ts#L234-L240)—I wonder whether the Homebridge accessories for thermostats and temperature sensors should be linked.

api/ring-types.ts
```diff
@@ -179,7 +181,7 @@ export interface RingDeviceData {
   locked?: 'jammed' | 'locked' | 'unlocked' | 'unknown'
   roomId?: number
   volume?: number
-  mode?: AlarmMode
+  mode?: AlarmMode | ThermostatMode
```
smockle, line 184: Outside the scope of this PR, but I briefly explored making `RingDeviceData` a [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions).

It’d allow us to mark fewer properties as optional (e.g. whenever `deviceType` is `temperature-control.thermostat`, then `setPoint` will exist). And it would allow us to clean-up `mode` (since no device actually supports all of `AlarmMode` and `ThermostatMode` values for `mode`).

I think we’d need to discriminate based on both `deviceType` and `categoryId`, in order to differentiate between e.g. `MultiLevelSwitch`, `MultiLevelBulb` and `Fan`.

homebridge/thermostat.ts
dgreif: I'm hesitant to add the component devices to the flat data of the parent.  The downside is you don't have a reference to the `RingDevice` instance for that component device, just to its data.  It may be more useful to add a `getComponentDevices(parentZid: string) => RingDevice[]` method to each device.  I haven't seen the `addLinkedService` method before but it definitely seems useful in this case.  Maybe after all accessories have been created in the platform, we could then go back and do linking.

api/ring-types.ts
dgreif: I agree that a discriminated union would be completely appropriate here.  The main difficulty in creating it will be getting accurate representations of the types for each individual accessory type.  Another problem is how to handle unknown device types since we don't know what their data will look like.  When I initially made this type, I started down that road but switched away just to save time and get something out the door.  Now could be a good time to re-address it in a separate PR.

homebridge/thermostat.ts
dgreif: Man, GitHub is being really odd about how it links my responses to your comments today 😕 .  Might want to have someone look into that 😉.  Anyways, `registerObservableCharacteristic` is part of the `BaseAccessory` and is intended to be used when you already have an `Observable` that is producing the exact values that you want to use for `Characteristic` updates.  It is mainly used by non-`RingDevice`s, like the camera and location Modes.  `registerCharacteristic` is on `BaseDataAccessory` and is very useful when you have a `RingDevice` and want to map from the `onData` events to the characteristic value.  In general, if you have a `RingDevice`, you should use `registerCharacteristic`.  The reason I did the `TemperatureSensor` and a few others differently is because I wanted to have a subscription to the `Observable` that would log when the sensor was faulted.  At that point I already had the `Observable` set up, so no need to do the `onData` mapping again by using `registerCharacteristic`.  Now that I look at it, I probably would have been better off adding a `handleUpdate` option to `registerCharacteristic` that would get called each time the characteristic changed value and could be used specifically for logging the events that we care about.

smockle:
Corrected the getters and settings for several characteristics in https://github.com/dgreif/ring/pull/331/commits/866d74b03ccad670eed291bc5462e04b32d1eef5, https://github.com/dgreif/ring/pull/331/commits/ef86cd5242d36f5b161e0ca47426a80a08997727 and https://github.com/dgreif/ring/pull/331/commits/968644cc5d2517f4118b51c1a8979634d40b33eb. Specifically, there was a copy-paste issue between `CurrentHeatingCoolingState` and `TargetHeatingCoolingState` (neither actually worked) and an invalid argument error (numbers instead of strings) in `TargetHeatingCoolingState`.

Using `npm run test-homebridge`, I tested two thermostats in the Home app—details have been added to the [first post](https://github.com/dgreif/ring/pull/331#issue-421681120) of this PR (copied below):
> - My thermostat appears in the Home app (along with correct current temperature, set point, mode, and status)
> - When I change set point or mode in the Home app, the physical unit reflects the changes
> - When I change set point or mode using the physical unit, the Home app reflects the changes

There is one outstanding issue I’d like address (though it’s not necessarily a blocker): “Auto” is displayed in the mode list in the Home app, but my thermostat does not support it. I’m not sure how to prevent it from showing up—the docs for the [`Thermostat` service](https://developers.homebridge.io/#/service/Thermostat) don’t mention a “supported modes” characteristic.

api/location.ts
smockle: Copy-pasting @dgreif’s [comment above](https://github.com/dgreif/ring/pull/331#discussion_r429253567):
> I'm hesitant to add the component devices to the flat data of the parent. The downside is you don't have a reference to the `RingDevice` instance for that component device, just to its data. It may be more useful to add a `getComponentDevices(parentZid: string) => RingDevice[]` method to each device. I haven't seen the `addLinkedService` method before but it definitely seems useful in this case. Maybe after all accessories have been created in the platform, we could then go back and do linking.

smockle: Added a `getComponentDevices` method in https://github.com/dgreif/ring/pull/331/commits/28df3881bbacf257e092b2d4d19d880830d8185b. I haven’t tested it as thoroughly as the previous implementation—I’ll try to do so tomorrow as temperatures fluctuate.

dgreif:
I ran into a similar issue with the doorbell buttons where I wanted to hide some of the possible outputs in HomeKit. Luckily it looks like Auto is the highest value for that characteristic, so you can set a max value like I did here: https://github.com/dgreif/ring/blob/d1cf0a5d3eae4d506d7aea839b4bb73140a042a0/homebridge/camera.ts#L125

I’ll do a full review of the PR tomorrow or Sunday. I really appreciate your work! It’s really nice to have another person with there mind this deep in the code, especially with your TypeScript experience. Thank you!

homebridge/thermostat.ts
```diff
@@ -0,0 +1,178 @@
...
+        // noop
+        // Setting display unit is unsupported
+      },
+    })
+  }
+
+  async getTemperatureFromComponentDevices(): Promise<number | undefined> {
+    const componentDevices = await this.device.getComponentDevices(),
+      temperatureSensorDevice = componentDevices?.find(({ data }) =>
+        data.deviceType.endsWith('sensor.temperature')
```
smockle, line 166: Just to double-check (because I’m a bit confused by the observables/how updates happen): When a component temperature sensor’s `data` is updated, will this function re-run (and trigger updates to `CurrentTemperature` and `CurrentHeatingCoolingState`, the thermostat characteristics which call it)?

smockle:
@smockle [said](https://github.com/dgreif/ring/pull/331#issuecomment-632974800):
> There is one¹ outstanding issue I’d like address (though it’s not necessarily a blocker): “Auto” is displayed in the mode list in the Home app, but my thermostat does not support it. I’m not sure how to prevent it from showing up—the docs for the Thermostat service don’t mention a “supported modes” characteristic.

@dgreif [said](https://github.com/dgreif/ring/pull/331#issuecomment-632977772):
> I ran into a similar issue with the doorbell buttons where I wanted to hide some of the possible outputs in HomeKit. Luckily it looks like Auto is the highest value for that characteristic, so you can set a max value like I did here: https://github.com/dgreif/ring/blob/d1cf0a5d3eae4d506d7aea839b4bb73140a042a0/homebridge/camera.ts#L125

Thanks for the pointer! Added a commit https://github.com/dgreif/ring/pull/331/commits/fac1b2594f6457b23bd6abe3a69a91406c228879 following the same pattern, and “Auto” is now hidden. (I used `validValues` instead of `maxValue` in case Homebridge gains support for additional modes (e.g. auxiliary/emergency heat) that we want to selectively support.)

homebridge/thermostat.ts
```diff
@@ -0,0 +1,189 @@
...
+      characteristicType: Characteristic.TemperatureDisplayUnits,
+      serviceType: Service.Thermostat,
+      getValue: () => {
+        // Neither thermostats nor their component devices (e.g. temperature sensors)
+        // appear to include the unit preference. Hardcoding Fahrenheit as the default.
+        return Characteristic.TemperatureDisplayUnits.FAHRENHEIT
+      },
+      setValue: () => {
+        // noop
+        // Setting display unit is unsupported
```
dgreif, line 169: Might be better to limit the `validValues` here since we aren't supporting Celcius

homebridge/thermostat.ts
```diff
@@ -0,0 +1,189 @@
...
+        if (
+          currentTemperatureIsLowerThanTarget &&
+          (mode === 'heat' || mode === 'aux')
+        ) {
+          // The current temperature is lower than the target temperature,
+          // and the thermostat is set to 'heat' or 'aux' (emergency heat), so the thermostat is heating
+          return Characteristic.CurrentHeatingCoolingState.HEAT
+        }
+        // The current temperature is either higher or lower than the target temperature,
+        // but the current thermostat mode would only to increase the difference,
```
dgreif, line 59:
```suggestion
        // but the current thermostat mode would only increase the difference,
```

homebridge/thermostat.ts
```diff
@@ -0,0 +1,189 @@
...
+    this.registerCharacteristic({
+      characteristicType: Characteristic.CurrentTemperature,
+      serviceType: Service.Thermostat,
+      getValue: async () => {
+        const temperature = await this.getTemperatureFromComponentDevices()
+        if (!temperature) {
+          return
+        }
+        // Documentation: https://developers.homebridge.io/#/characteristic/CurrentTemperature
+        // 'Characteristic.CurrentTemperature' supports 0.1 increments
```
dgreif, line 126: https://github.com/chrisjshull/homebridge-nest/blob/master/lib/nest-thermostat-accessory.js#L70 - looks like you _can_ adjust this step size, though I'm not sure what step size ring supports so `0.1` is probably just fine.

homebridge/thermostat.ts
```diff
@@ -0,0 +1,189 @@
...
+        return Math.round(temperature * 10) / 10
```
dgreif, line 127: Slightly different syntax option, if you think it's more clear:

```suggestion
        return Number(temperature.toFixed(1))
```

homebridge/thermostat.ts
```diff
@@ -0,0 +1,189 @@
...
+      setValue: (setPoint: number) => {
+        this.logger.info(
+          `Setting ${this.device.name} target temperature to ${setPoint}`
+        )
+
+        // Documentation: https://developers.homebridge.io/#/characteristic/TargetTemperature
+        // 'Characteristic.TargetTemperature' has a valid range from 10 to 38 degrees celsius,
+        // but devices may support a different range. When limits differ, accept the more strict.
+        const setPointMin = Math.max(this.device.data.setPointMin || 10, 10),
+          setPointMax = Math.min(this.device.data.setPointMax || 38, 38)
```
dgreif, line 146: Thoughts on overriding the minValue and maxValue on the `TargetTemperature` if they are present on the device data?

homebridge/thermostat.ts
dgreif: I can definitely see where this would be confusing, especially now with component devices in the mix.  Each `RingDevice` has an `onData` observable that will fire whenever the data for that particular device is updated.  Each fires once with the initial data, and then again when we receive websocket messages which are device specific.  `registerCharacteristic` is specifically designed around using the `onData` observable attached to the device used to create the accessory.  This means it will _not_ be listening for changes to the data on the component devices.  So if the temperature sensor updates its data, the parent thermostat will not fire a new event from its `onData`.  In this case, it would be best to grab the child temp sensor `RingDevice` and then map its `onData` observable down to a temperature.  With that observable, you can then use the `registerObservableCharacteristic` instead of `registerCharacteristic`.  Let me know if this is too convoluted, in which case I can just merge what you have and make the changes.

homebridge/thermostat.ts
smockle: I considered that, but `toFixed` throws a SyntaxError when `temperature` is an integer:
```
> 21.toFixed(1)
21.toFixed(1)
^^^

Uncaught SyntaxError: Invalid or unexpected token
```

I do think it’s more clear, and it looks like this slight variation works:
```JS
Number(Number(21).toFixed(1))
```

smockle: Committed in https://github.com/dgreif/ring/pull/331/commits/52123c148ec0370a94c71168a9cc96520d15812b

homebridge/thermostat.ts
smockle: Oh interesting, thanks for the tip! I wonder whether the issue alluded to in [the comment](https://github.com/chrisjshull/homebridge-nest/blob/a2165e5b572dfa488408ef09192099852d2619d9/lib/nest-thermostat-accessory.js#L59-L62) a few lines above will affect us:
```JS
/*
 * Only allow 0.5 increments for Celsius temperatures. HomeKit is already limited to 1-degree increments in Fahrenheit,
 * and setting this value for Fahrenheit will cause HomeKit to incorrectly round values when converting from °F to °C and back.
 */
```

A `0.1` degrees celsius step is about `0.18` degrees Fahrenheit at typical room temperatures. The Home app only supports round-number set points. I wouldn’t be surprised if it supported half-degree set points in the future, but I doubt it ever supports anything finer than that (e.g. `0.18`).

> I'm not sure what step size ring supports so 0.1 is probably just fine.

I also don’t know. I propose we leave the step as-is for now, and revisit in the future if someone reports a disparity between the Fahrenheit number displayed on their wall unit and the Fahrenheit number displayed in the Home app (the latter converted from the celsius value included in the API response).

homebridge/thermostat.ts
smockle: Thanks! Committed in https://github.com/dgreif/ring/pull/331/commits/8559f7f44a73437ec23a1c4070dce19b6ea8f651

homebridge/thermostat.ts
smockle: I attempted to limit the values in https://github.com/dgreif/ring/pull/331/commits/7d8eca5362c2c69e04ed24a667513931e2d4aff0, but both display unit values are still selectable in the Home app:

IMG_3114

homebridge/thermostat.ts
smockle: 👍 Good idea! Added in https://github.com/dgreif/ring/pull/331/commits/ce3072ac677d7aee2e0529c278476262204a4bb6

homebridge/thermostat.ts
smockle: > In this case, it would be best to grab the child temp sensor `RingDevice` and then map its `onData` observable down to a temperature. With that observable, you can then use the `registerObservableCharacteristic` instead of `registerCharacteristic`.

I attempted this in https://github.com/dgreif/ring/pull/331/commits/768154bfc9987d2cbc336a21c768fe07c3881b99, but it doesn’t look like the `onValue` methods are getting called, so I’ve missed something. I’ll keep working on it; if you have any suggestions though, please let me know!

smockle: I think this is the simplest case to illustrate what I don’t understand:
```JS
location.onDevices
.pipe(
  filter((device) => device.initialData.parentZid === this.id || device.data.parentZid === this.id)
)
```
This code does not compile: I’m expecting `device` (`filter` argument) to be a `RingDevice`, but instead it’s a `RingDevice[]`.

smockle: I think I may have figured it out! I see temperature readings from the component sensors using https://github.com/dgreif/ring/pull/331/commits/e91c6312c4590c735d18964c99fe96a3b81f9cac

api/ring-device.ts
```diff
@@ -32,15 +32,13 @@ export class RingDevice {
...
-        return acc.concat(componentDevices || [])
-      }, [] as RingDevice[])
-    )
+  getComponentDevices(callback: (componentDevices: RingDevice[]) => void) {
+    return this.location.onDevices.subscribe((devices) => {
+      const componentDevices = devices.filter(
+        ({ data }) => data.parentZid === this.id
+      )
+      callback(componentDevices)
+    })
```
dgreif, line 41: If you want to go with `getComponentDevices` instead of the `onComponentDevices` observable (which is totally fine), I'd prefer that it use an async function instead of a callback.  

```suggestion
  async getComponentDevices(callback: (componentDevices: RingDevice[]) => void) {
    const devices = await this.location.getDevices()
    return componentDevices = devices.filter(
        ({ data }) => data.parentZid === this.id
      )
```

smockle: Makes sense, but I switched back to an `Observable`-returning `onComponentDevices`. The `Observable`-based API seemed more consistent with the classes, and could be invoked easily in the `Thermostat` constructor (no need for a `;(async () => {})()`-style [IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)).

homebridge/thermostat.ts
```diff
@@ -0,0 +1,237 @@
...
+    )
+
+    // Required Characteristics
+
+    this.registerObservableCharacteristic({
+      characteristicType: Characteristic.CurrentHeatingCoolingState,
+      serviceType: Service.Thermostat,
+      onValue: this.onTemperature.pipe(
+        map((temperature) => {
+          const { setPoint, mode } = this.device.data
```
smockle, line 56: Spotted a regression—`CurrentHeatingCoolingState` is re-evaluated when `this.device.data.setPoint` or `this.device.data.mode` change.

api/ring-device.ts
```diff
@@ -32,6 +32,12 @@ export class RingDevice {
     return this.data.name
   }
 
+  get onComponentDevices() {
+    return this.location.onDevices.pipe(
+      map((devices) => devices.filter(({ data }) => data.parentZid === this.id))
+    )
+  }
```
dgreif, line 39: Shouldn't need the getter here.  Can just create an observable directly as a member.  The nice thinng is that they don't actually execute unless they are subscribed to, so it's no additional load if you don't use a getter.

homebridge/thermostat.ts
dgreif, line 56: This should fix your regression.  CombineLatest will fire an event whenever any of the observables fire (once all have fired at lease once)

```suggestion
      onValue: combineLatest([this.onTemperature, this.device.onData]).pipe(
        map(([temperature, { setPoint, mode }])) => {
```

api/ring-device.ts
smockle: Nice catch! Fixed in https://github.com/dgreif/ring/pull/331/commits/697fb3244b860642dd92f86aade3adaac3e5ec7f

homebridge/thermostat.ts
smockle: Resolved by commits https://github.com/dgreif/ring/pull/331/commits/bad31ecc23e34f9d44b3f1bef4983d79c9828961 and https://github.com/dgreif/ring/pull/331/commits/bca661528787802e0174fa1f37af7fb39ce92079

smockle: Verified this works as expected. 👍

PR #341, opened 2020-06-01
Overall reduction in network overhead and latency

DragonWork:
- Removed axios and added the pure https client.
- Added the dnscache module to reduce the frequency of dns requests to the same domain within a short period of time.
- Tremendeously reduces network overhead and roundtrip time by keeping TCP sockets alive, so successive longpolling is actually made on the previous existing connection.

### DIFF
--- a/api/dnscache.d.ts
+++ b/api/dnscache.d.ts
@@ -1,0 +1,1 @@
+declare module 'dnscache';
--- a/api/rest-client.ts
+++ b/api/rest-client.ts
@@ -1,8 +1,15 @@
-import axios, { AxiosRequestConfig, ResponseType } from 'axios'
+import https from 'https'
+import url from 'url'
+import dnscache from 'dnscache'
 import { delay, getHardwareId, logError, logInfo, stringify } from './util'
 import * as querystring from 'querystring'
 import { AuthTokenResponse, SessionResponse } from './ring-types'
 import { ReplaySubject } from 'rxjs'
+
+const httpsAgent = new https.Agent({ keepAlive: true });
+https.globalAgent = httpsAgent;
+
+const dnsCacheLookup = dnscache({ enabled: true, ttl: 60 }).lookup;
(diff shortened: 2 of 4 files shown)


dgreif:
@DragonWork  thank you very much for this PR! It’s perfect timing as this was the last thing I had wanted to change before releasing version 9. I was planning to switch from axios to ‘got’, which has dnscache built in. I hadn’t considered keepalive yet, but that is a fantastic idea, and completely applicable for this use case. I’m a little hesitant to use base https since there are lots of gotchas (like the header casing, which you already accounted for) that you avoid by using one of the other node modules. I’m not hard set either way, but what do you think about using got?

DragonWork:
Initially, I wanted to use HTTP/2 (as I'm doing privately for communicating with the Alexa's Smart Home API), which supports multiplexing concurrent requests into a single robust connection. HTTP/2 is definitely supported by Amazon's infrastructure, but Ring seems to use AWS only at the _app.ring.com_ endpoint, _api.ring.com_ (which is requested more often) oddly uses CloudFront as a proxy. CloudFront wants their customers to explicitly allow HTTP/2 to work, and Ring hasn't done yet. Well, this might change in the future. But I didn't want to add node.js' vanilla HTTP/2 module just to talk to _app.ring.com_.

**got** seems to actually provide HTTP/2 under the hood, if it's supported – that would be a plus too. So I totally agree to use **got** instead of my error-prone quick hack.

PS: I never used TypeScript before, and naturally got a ton of errors until it I figured the basics out. Feel free to adjust my code as needed!

dgreif:
@DragonWork thanks for the feedback.  I agree it would be great to use HTTP/2 where applicable, and if `got` does that automatically, the I definitely think that's the way to go.  I think it will be easiest to close out this PR and I'll just make the changes on my end.  Once it's out, I'd love your feedback on perf differences!

PR #342, opened 2020-06-01
Second attempt. ;-)

DragonWork:
I tried to implement your feedback, offloading the http logic to **got** and removing **dnscache** in favor of the **cacheable-lookup** module, which comes bundled with **got**. Give it a try.

### DIFF
--- a/api/location.ts
+++ b/api/location.ts
@@ -309,15 +309,14 @@ export class Location {
 
   setLightGroup(groupId: string, on: boolean, durationSeconds = 60) {
     this.restClient.request<any>({
-      method: 'POST',
+      method: 'post',
       url: `https://api.ring.com/groups/v1/locations/${this.id}/groups/${groupId}/devices`,
-      data: {
+      json: {
         lights_on: {
           duration_seconds: durationSeconds,
           enabled: on,
         },
       },
-      json: true,
     })
   }
 
@@ -427,12 +426,11 @@ export class Location {
     }
 
     return this.restClient.request<AccountMonitoringStatus>({
-      method: 'POST',
+      method: 'post',
       url: appApi(
         `rs/monitoring/accounts/${this.id}/assets/${baseStationAsset.uuid}/userAlarm`
       ),
-      json: true,
-      data: {
+      json: {
         alarmSessionUuid,
         currentTsMs: now,
         eventOccurredTime: now,
@@ -453,7 +451,7 @@ export class Location {
     this.onLocationModeRequested.next()
 
     const response = await this.restClient.request<LocationModeResponse>({
-      method: 'GET',
+      method: 'get',
       url: appApi(`mode/location/${this.id}`),
     })
 
@@ -464,10 +462,9 @@ export class Location {
 
   async setLocationMode(mode: LocationModeInput) {
     const response = await this.restClient.request<LocationModeResponse>({
-      method: 'POST',
+      method: 'post',
       url: appApi(`mode/location/${this.id}`),
-      json: true,
-      data: { mode },
+      json: { mode },
     })
 
     this.onLocationMode.next(response.mode)
@@ -477,7 +474,7 @@ export class Location {
 
   async disableLocationModes() {
     await this.restClient.request<void>({
-      method: 'DELETE',
+      method: 'delete',
       url: appApi(`mode/location/${this.id}/settings`),
     })
     this.onLocationMode.next('disabled')
@@ -485,32 +482,31 @@ export class Location {
 
   getLocationModeSettings() {
     return this.restClient.request<LocationModeSettingsResponse>({
-      method: 'GET',
+      method: 'get',
       url: appApi(`mode/location/${this.id}/settings`),
     })
   }
 
   setLocationModeSettings(settings: LocationModeSettings) {
     return this.restClient.request<LocationModeSettingsResponse>({
-      method: 'POST',
+      method: 'post',
       url: appApi(`mode/location/${this.id}/settings`),
-      data: settings,
+      json: settings,
     })
   }
 
   getLocationModeSharing() {
     return this.restClient.request<LocationModeSharing>({
-      method: 'GET',
+      method: 'get',
       url: appApi(`mode/location/${this.id}/sharing`),
     })
   }
 
   setLocationModeSharing(sharedUsersEnabled: boolean) {
     return this.restClient.request<LocationModeSharing>({
-      method: 'POST',
+      method: 'post',
       url: appApi(`mode/location/${this.id}/sharing`),
-      json: true,
-      data: { sharedUsersEnabled },
+      json: { sharedUsersEnabled },
     })
   }
 
--- a/api/location.ts
+++ b/api/location.ts
@@ -309,7 +309,7 @@ export class Location {
 
   setLightGroup(groupId: string, on: boolean, durationSeconds = 60) {
     this.restClient.request<any>({
-      method: 'post',
+      method: 'POST',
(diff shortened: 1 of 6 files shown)


DragonWork:
PS: View the code by comparing the changes against your master. Otherwise things may get unnecessarily complicated.

api/location.ts
```diff
@@ -309,15 +309,14 @@ export class Location {
 
   setLightGroup(groupId: string, on: boolean, durationSeconds = 60) {
     this.restClient.request<any>({
-      method: 'POST',
+      method: 'post',
```
dgreif, line 312: Looks like it works just fine if you leave all of these as `POST`/`GET` etc, which should help cut down on the size of your PR

api/rest-client.ts
```diff
@@ -1,9 +1,24 @@
...
+import got, { NormalizedOptions, Options, Headers } from 'got'
+import CacheableLookup from 'cacheable-lookup'
 import { delay, getHardwareId, logError, logInfo, stringify } from './util'
 import * as querystring from 'querystring'
 import { AuthTokenResponse, SessionResponse } from './ring-types'
 import { ReplaySubject } from 'rxjs'
 
+const cacheableLookup = new CacheableLookup();
+const httpsAgent = new https.Agent({ keepAlive: true });
+https.globalAgent = httpsAgent;
```
dgreif, line 11: Shouldn't need to set this anymore.  Setting `globalAgent` seems like it could have an impact on other homebridge modules as well, so is something I wanted to avoid.  I'd prefer to use `keepaliveagent`, as shown here: https://www.npmjs.com/package/got#agent

api/rest-client.ts
```diff
@@ -29,25 +44,24 @@ export interface ExtendedResponse {
...
   try {
-    const { data, headers } = await axios(options)
-
-    if (typeof data === 'object' && headers.date) {
-      data.responseTimestamp = new Date(headers.date).getTime()
+  	const options = { ...defaultOptions, ...reqOptions };
+    const { headers, body } = <{ headers: Headers, body: any }>await got(options);
+	const data = body as T & ExtendedResponse;
+    if (data !== null && typeof data === 'object' && headers.date) {
+      data!.responseTimestamp = new Date(<string>headers.date).getTime()
```
dgreif, line 54: Should be able to remove the `!` since you are checking for null
```suggestion
      data.responseTimestamp = new Date(<string>headers.date).getTime()
```

api/rest-client.ts
```diff
@@ -108,17 +122,16 @@ export class RingRestClient {
...
           ...grantData,
         },
-        method: 'POST',
+        method: 'post',
         headers: {
-          'content-type': 'application/json',
           '2fa-support': 'true',
           '2fa-code': twoFactorAuthCode || '',
-          hardware_id: await hardwareIdPromise,
+          'hardware_id': await hardwareIdPromise,
```
dgreif, line 134: This will probably cause a lint error.  You can check by running `npm run lint`

api/rest-client.ts
```diff
@@ -1,9 +1,24 @@
...
 import * as querystring from 'querystring'
 import { AuthTokenResponse, SessionResponse } from './ring-types'
 import { ReplaySubject } from 'rxjs'
 
+const cacheableLookup = new CacheableLookup();
+const httpsAgent = new https.Agent({ keepAlive: true });
+https.globalAgent = httpsAgent;
+
+const defaultOptions : Options = {
+  headers: { 'user-agent': 'Mozilla/5.0 (Linux) Ring' },
```
dgreif, line 14: Not sure where this user agent came from...

api/rest-client.ts
```diff
@@ -285,29 +285,27 @@ export class RingRestClient {
...
-          )}).`
+          `http request failed.  ${url} returned unknown errors: (${stringify(errors)}).`
         )
       }
 
-      if (response.status === 404 && url.startsWith(clientApiBaseUrl)) {
+      if (response.statusCode === 404 && url.startsWith(clientApiBaseUrl)) {
         logError('404 from endpoint ' + url)
-        if (response.data?.error?.includes(hardwareId)) {
+        if (response.body?.error?.includes(hardwareId)) {
```
dgreif, line 294: I'm not sure if this is the correct logic anymore, but something I'll have to test on my end

package.json
```diff
@@ -69,8 +69,9 @@
   },
   "devDependencies": {
     "@types/debug": "^4.1.5",
+    "@types/got": "^9.6.11",
```
dgreif, line 72: `got` ships with its own type definitions.  Was there a reason you had to pull these in?

api/ring-camera.ts
```diff
@@ -245,10 +245,16 @@ export class RingCamera {
...
-      data: {
-        'doorbot[settings][chime_settings][enable]': on,
+      form: {
+        doorbot: {
+        	settings: {
+        		chime_settings: {
+        			enable: 'on',
+        		},
+        	},
+        },
```
dgreif, line 257: Nice! Way better than what I had before

dgreif:
@DragonWork Any chance you will get to this today?  I'm fine with making the changes myself, just don't want to duplicate your efforts

api/location.ts
DragonWork: Actually it was the TypeScript compiler that complained about the uppercase method names...

api/rest-client.ts
DragonWork: Was trying to resolve a 400 Bad Request error, probably didn't help at all. Shouldn't we add something though? Like plain "Device", as I read some of the communication excerpts with the Ring servers.

api/rest-client.ts
DragonWork: I thought it is good practice to quote all JSON keys according to the standard? It isn't representing a property for the purpose of OOP after all. Although I can remove the quotes, that's not an issue.

api/rest-client.ts
DragonWork: When I was debugging that part, e.reponse was actually set. (It is thrown from the **got** method in **requestWithRetry**, where the try/catch just forwards the error object when it contains the e.response property (__throw e__ in catch). If not, another error is being logged and a 5s-delayed retry is set up.

package.json
DragonWork: Like I said before, that's my first try with TypeScript. Just thought it wouldn't compile like when I added the base https module. :) I'm removing it.

api/ring-camera.ts
DragonWork: Actually there is a small bug in here. The 'on' argument of the function is not respected, as I wrongly typed 'on' instead of on. 🤦

api/rest-client.ts
DragonWork: True that!

api/rest-client.ts
dgreif: JSON does require that all keys be quoted, but javascript objects to not require it.  In plain javascript objects it works with or without quotes, unless you have special characters in the key (like the `-` in the keys above).  Most javascript linters will have quotes removed when possible since they unnecessary extra characters in this case.

api/rest-client.ts
dgreif: Not a big deal either way.  They aren't using `user-agent` to do any verification on their end, so no need to provide it.

dgreif:
@DragonWork no worries about the delay!  Just wanted to see if you planned to get to it today or if I should take it from where you had it.  I really appreciate you taking the time to make changes!  I'm going to merge this into a branch for local testing, and then roll it into the next v9 beta release.  Should be out tomorrow or the next day.  Thanks for all your hard work!

PR #361, opened 2020-06-14
Update return-audio-example.ts

adenta:


examples/return-audio-example.ts
```diff
@@ -7,7 +7,7 @@ import { decodeCryptoValue, getSrtpValue, RtpSplitter } from '../api/rtp-utils'
 import { take } from 'rxjs/operators'
 
 /**
- * This example records a 10 second video clip to output/example.mp4
+ * This example takes an audio clip from output/example.mp4 and pipes it to a ring device
```
dgreif, line 10:
```suggestion
 * This example takes an audio clip from examples/example.mp4 and pipes it to a ring camera
```

dgreif:
Just curious, what interface are you piping all the video and audio to/from?  It's helpful to know how it's being used so that I can make it easier and ensure I don't release any breaking changes

adenta:
@dgreif I'm running your library on heroku, and ingesting it in my own website. I've got it set up to send a pre-recorded audio file. Def going to try and hack it together to send audio from my website, directly to the ring.

Biggest problem I have right now is lag in the video feed, unsure what all I can do to get the latency down to like, 1-2 seconds like the official ring app.

dgreif:
Sounds good.  I haven't done any work on streaming directly to the web browser, but I do know that Ring is supporting WebRTC for amazon fire tv and echo devices.  It will take more exploring, but it's possible that we could use that interface instead of SIP to initiate the stream, which would work in the browser.  I'm not planning to look into this any time soon, but it might be possible theoretically

adenta:
would love to stay in contact, feel free to reach out at https://news.ycombinator.com/user?id=adenta

Issue #167, opened 2019-11-01
InHomeDoorbellSwitch not working with Chimes


Tangston311:
First, this is an incredible plugin!  Well done.  

I was excited to see the addition of the InHomeDoorbellSwitch function so I could automate turning my doorbell on or off, but I've found that regardless of the state of the InHomeDoorbellSwitch switch my doorbell still always rings...I have the Elite Video Doorbell; is that one of the doorbells where this switch won't work, or is something else going on?

I'm a little out of touch on Homebridge lately so let me know what other information and/or logs I can provide that might be helpful.

dgreif:
Sorry to hear you are having issues.  Please upgrade to v`5.9.1` if you haven't already and give it another try.  If you are just looking at the Ring app to see if it updated, you have to go back to the main screen and then into settings for the camera settings to show the correct value.  Let me know if you are for-sure having issues after checking all of that.

Tangston311:
Ok I managed to dig my hole a little deeper - I updated to 5.9.1 and did as you suggested in the Ring app, but the switch still had no effect.  So I tried deleting the Ring Camera from my Home app with the intent of re-adding it to see if that would help, but now I can't get it to appear when I select "Add a new Accessory" and "I don't have a code or can't scan".  I tried deleting and re-installing the plugin entirely, and removing and re-adding Homebridge, but I can't get the doorbell to re-appear.

Any suggestions?  Thanks!

dgreif:
https://github.com/dgreif/ring/wiki/Camera-Troubleshooting#cannot-add-camera-from-home

Tangston311:
Ok that got the camera to show back up! Thank you - great documentation by the way, I should have checked around for that.

Unfortunately turning off or on the doorbell switch still has no effect on the operation of the doorbell (either by looking within the Ring app itself or by testing the doorbell by ringing it).

dgreif:
Is the in home doorbell stuck on or off?  Does it toggle if you change the setting in the Ring app? @smockle has it been working for you?

Tangston311:
It's not stuck - I can toggle it off and on in the Home app, it just doesn't appear to have any influence on the behavior of the doorbell.  I'm not sure exactly which setting in the Ring app I should be checking, but I tried toggling the switch in the Home app to see if it affected any of the following settings in the Ring app, and none of them did:

- Ring Alerts (set to "On" in the Ring app, no change when switch is toggled in Home app)
[...]

PR #365, opened 2020-06-17
Customizable Chime snooze duration via Brightness setting

dxdc:
"Brightness" is defined as a % of a day. The only allowable values are in discrete hours (0-24).

See discussion in: https://github.com/dgreif/ring/issues/167 regarding the benefits of this approach. This implementation is working.

To install this, it requires deleting the existing Chime. I do this as follows:
```sh
cat cachedAccessories | jq 'del(.[] | select(.displayName|test("Chime")))' -c > cachedAccessories.new
mv cachedAccessories.new cachedAccessories
```

I realize that `Brightness` may not be the ideal characteristic. Maybe using degrees (e.g., for a thermostat) would be more sensible, but I'd have to think through how to best do this given that users can have their settings in deg. C or F.

Also, we could think through the best icon to use, although users can still edit the lightbulb icon to another one of their choosing.

### DIFF
--- a/homebridge/chime.ts
+++ b/homebridge/chime.ts
@@ -5,8 +5,6 @@ import { Logging, PlatformAccessory } from 'homebridge'
 import { BaseDataAccessory } from './base-data-accessory'
 import { logInfo } from '../api/util'
 
-const minutesFor24Hours = 24 * 60
-
 export class Chime extends BaseDataAccessory<RingChime> {
   constructor(
     public readonly device: RingChime,
--- a/homebridge/chime.ts
+++ b/homebridge/chime.ts
@@ -17,7 +17,7 @@ export class Chime extends BaseDataAccessory<RingChime> {
     super()
     const { Characteristic, Service } = hap,
       snoozeService = this.getService(
-        Service.Switch,
+        Service.Lightbulb,
         device.name + ' Snooze',
         'snooze'
       ),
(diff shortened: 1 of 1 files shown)


dgreif:
What about a valve?  I has duration features built in: https://developers.homebridge.io/#/service/Valve

dxdc:
> What about a valve? I has duration features built in:

I thought about it. But, the 1 hour restriction (3600 sec) seemed unappealing. Also, it counts down automatically from recollection when I played with this before. A little confusing, maybe?

```
export class SetDuration extends Characteristic {

  static readonly UUID: string = '000000D3-0000-1000-8000-0026BB765291';

  constructor() {
    super('Set Duration', SetDuration.UUID);
    this.setProps({
      format: Formats.UINT32,
      maxValue: 3600,
      minValue: 0,
      minStep: 1,
      perms: [Perms.READ, Perms.WRITE, Perms.NOTIFY]
    });
    this.value = this.getDefaultValue();
  }
}
```

dgreif:
We should be able to increase the max by calling `setProps` on the characteristic, bumping it up to 24 hours instead.  There are a few examples already in this repo using `setProps` if you need a reference.  I think counting down would be handy, especially since Ring does that on their end as well.  We would need to look at `device.data.do_not_disturb.seconds_left` for an accurate idea of what Ring says is left.

dxdc:
You can increase `setProps` past default max values?? Didn't test this before, but if so, I agree it could be very interesting. Easy enough to try!

dxdc:
@dgreif OK... refactored this using Valve. I put it in a different branch in case you want to test both. Thanks for the tip that you can exceed `maxValue`! I didn't know that before.

dgreif:
I think the valve will be the better solution in the end.  Closing this PR in favor of #370

PR #370, opened 2020-06-17
Chime snooze using valve

dxdc:
Different implementation of https://github.com/dgreif/ring/pull/365 using `Valve` instead of `Brightness`

Still there are some notes/quirks:
- Use of faucet icon probably isn't ideal...
- I have set the default on/off to 24 hours as previous, but user can go into accessory to customize the actual time
- Not sure if `SetDuration` requires `get`?

### DIFF
--- a/homebridge/chime.ts
+++ b/homebridge/chime.ts
@@ -5,7 +5,13 @@ import { Logging, PlatformAccessory } from 'homebridge'
 import { BaseDataAccessory } from './base-data-accessory'
 import { logInfo } from '../api/util'
 
-const minutesFor24Hours = 24 * 60
+const secondsFor24Hours = 60 * 60 * 24
+const elapsedTime = (elapsed: number) => {
+    const hours = Math.floor(elapsed / 3600),
+        minutes = Math.floor(elapsed / 60) % 60,
+        seconds = elapsed % 60
+    return `${hours}:${minutes.toString().padStart(2, '0')}:${Math.round(seconds).toString().padStart(2, '0')}`
+}
 
 export class Chime extends BaseDataAccessory<RingChime> {
   constructor(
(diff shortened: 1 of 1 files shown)


dgreif:
I'll need to do some additional testing and tweaking.  I'm planning to have this be an option so that snooze shows as a switch by default, but users can add in the valve.  That way they aren't seeing some random water valve without at least some idea of what they are getting into and why.  Thanks for getting this started, looks like it will work great!

dgreif:
@dxdc sorry for the super long delay on this.  I started pulling it in this morning and realized that valves apparently can't be automated 😕.  I have no idea why HomeKit has this limitation, but it seems to me that it defeats the point of changing this to a valve.  Can you double check that it doesn't allow you to add the valve to automations?  If so, is this still useful to you?

dxdc:
@dgreif ah, you're right. The built-in automations in HK are honestly so useless to me though. There is no way to organize/group them, name them, to set multiple "IF" commands, etc. I use NodeRED almost exclusively for managing automations, it is vastly superior in every way.

So, it is still useful for me to snooze chime alerts by a time interval because I can do so with NodeRED. Still I'm sure I'm in the minority for that one ;)

- What about reverting to the brightness idea. Less intuitive though.
- Or even the temperature control, using `HeatingThresholdTemperature`? I'm not sure how to implement that, can we implement it as a solo characteristic? And, `TemperatureDisplayUnits` can be fixed to Celsius?

I really wish there was another kind of numerical slider that was more useful though in HK.

dgreif:
Since the work is already some, I think I’ll just add this as an undocumented config option. That way you can use it and it’s there if others ask about it. Will probably get it released tomorrow morning

Issue #337, opened 2020-05-27
406 Not Acceptable error


loudgazelle:
# Bug Report

### Describe the Bug
Out of the blue I started getting a `406 Not Acceptable` error from homebridge-ring, similar to the issues described in #247 . Prior to receiving this error I had not changed any login information, but after I started seeing it I refreshed my token via the GUI in hombebridge-config-ui-x, and when that didn't resolve the error I refreshed my token via the process described [here on issue #247](https://github.com/dgreif/ring/issues/247#issuecomment-579264986). Neither refresh has resolved the error. 

This appears to have put home bridge in a restart loop, so I've stopped home bridge altogether so that I don't overload the APIs used by the other plugins. 

### To Reproduce
 - Start Homebridge
 - See error in logs

### Expected behavior
no error, plugin works as it was yesterday. 

### Screenshots/Logs
```
[5/27/2020, 12:19:08 PM] [Ring] Request to https://app.ring.com/rhq/v1/devices/v1/locations failed with status 406. Response body: <html>
<head><title>406 Not Acceptable</title></head>
<body bgcolor="white">
<center><h1>406 Not Acceptable</h1></center>
<hr><center>openresty</center>
</body>
</html>

[5/27/2020, 12:19:08 PM] [Ring] Error connecting to API
[5/27/2020, 12:19:08 PM] [Ring] Error: Request failed with status code 406
    at createError (/usr/local/lib/node_modules/homebridge-ring/node_modules/axios/lib/core/createError.js:16:15)
    at settle (/usr/local/lib/node_modules/homebridge-ring/node_modules/axios/lib/core/settle.js:17:12)
    at IncomingMessage.handleStreamEnd (/usr/local/lib/node_modules/homebridge-ring/node_modules/axios/lib/adapters/http.js:236:11)
    at IncomingMessage.emit (events.js:322:22)
    at endReadableNT (_stream_readable.js:1187:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  config: {
    url: 'https://api.ring.com/clients_api/session',
    method: 'post',
    data: '{"device":{"hardware_id":"XXXX","metadata":{"api_version":11,"device_model":"homebridge-ring"},"os":"android"}}',
    headers: {
      Accept: 'application/json, text/plain, */*',
      'Content-Type': 'application/json',
      authorization: 'Bearer XXXX',
      'User-Agent': 'axios/0.19.2',
      'Content-Length': 143
    },
[...]

Issue #247, opened 2020-01-27
406 Not Acceptable error


rpavlik:
# Bug Report

### Describe the Bug

I can't log in - I'm getting a "406 Not Acceptable" error. I changed my script-dedicated account password following an unexpected "got a login from an Android app" notification in my email (though looking at the debug, that was probably just this package). I also updated ring-alarm-mqtt (and thus this package) before putting the new (longer) password.

### To Reproduce

 - Run the auth-cli script - I've tried using the latest master as well as an older version that previously worked.
 - Attempt to log in (I'm using a 24-character password, which is the only thing I changed between it working and not working. Have tried down to a 12 character which still doesn't work.)
 - See error

### Expected behavior

Login to work - as I mentioned, I was using this before (older version, shorter password) successfully for quite some time.

### Screenshots/Logs

Not sure what parts of the logs are useful here - I've erred on the side of "include nothing that might be related to a token"... hopefully it's still useful.

```
  follow-redirects options { protocol: 'https:',
  maxRedirects: 21,
  maxBodyLength: 10485760,
  path: '/oauth/token',
  method: 'POST',
  headers:
   { Accept: 'application/json, text/plain, */*',
     'Content-Type': 'application/json',
     '2fa-support': 'true',
     '2fa-code': '',
     hardware_id: 'XXXX',
     'User-Agent': 'axios/0.19.1',
     'Content-Length': 151 },
  agent: undefined,
  agents: { http: undefined, https: undefined },
  auth: undefined,
  hostname: 'oauth.ring.com',
  port: null,
  nativeProtocols:
   { 'http:':
      { _connectionListener: [Function: connectionListener],
        METHODS: [Array],
        STATUS_CODES: [Object],
        Agent: [Function],
        ClientRequest: [Function],
        globalAgent: [Agent],
        IncomingMessage: [Function],
        OutgoingMessage: [Function],
        Server: [Function],
        ServerResponse: [Function],
        createServer: [Function: createServer],
        get: [Function: get],
        request: [Function: request],
        maxHeaderSize: [Getter] },
     'https:':
      { Agent: [Function],
        globalAgent: [Agent],
        Server: [Function],
        createServer: [Function: createServer],
        get: [Function: get],
        request: [Function: request] } } } +0ms
  ring { status: 406,
  ring   statusText: 'Not Acceptable',
  ring   headers:
  ring    { date: 'Mon, 27 Jan 2020 22:45:15 GMT',
  ring      'content-type': 'text/html',
[...]

Issue #378, opened 2020-06-26
Ability to disconnect RingApi in order to update options


djm181:
# Feature Request

### Description
Once a RingApi instance is set up with a polling interval and refresh token there is no way to have it stop polling or change any of the options associated with the instance. For example if the set of locations to be monitored changes, one cannot change the `options`.

### Describe the solution you'd like
Would like to be able to disconnect the RingApi object and dispose of it so it is no longer polling for dings or active in any way etc, then would simply create a new instance

### Describe alternatives you've considered
Considered restarting the process with new options, but this increases the complexity of the solution as the Ring portion of the application must be hosted in a separate process in order to do this.

dgreif:
@djm181 I think this is a completely valid request.  No promises on timeline as I have been a bit limited on free time lately.  I would be happy to take a PR if you want to dive into the api code.

dgreif:
`api.disconnect()` added in v`9.6.0`

PR #389, opened 2020-07-01
feat(api): enable retrieve/reuse sessions (#337)

djm181:
Allows for sessions to be retrieved and subsequently reused by the client. Ring limits session requests and this enables getting around the 406 errors that the API starts to throw.

If a single IP makes too many `/session` requests (I think the limit is maybe 50 in a 24-hour period) then you run into #337 and #247 errors. But the session response is fairly status so we can avoid the request if we know that the session is good.

### DIFF
--- a/api/rest-client.ts
+++ b/api/rest-client.ts
@@ -99,13 +99,14 @@ export interface RefreshTokenAuth {
 
 export interface SessionOptions {
   controlCenterDisplayName?: string
+  reuseSession?: SessionResponse
 }
 
 export class RingRestClient {
   // prettier-ignore
   public refreshToken = ('refreshToken' in this.authOptions ? this.authOptions.refreshToken : undefined)
   private authPromise = this.getAuth()
-  private sessionPromise = this.getSession()
+  private sessionPromise = (this.authOptions.reuseSession ? Promise.resolve(this.authOptions.reuseSession) : this.getSession())
   public using2fa = false
   public onRefreshTokenUpdated = new ReplaySubject<{
     oldRefreshToken?: string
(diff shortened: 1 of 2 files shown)


djm181:
@dgreif - first PR on this repo so please let me know if there is anything else you want me to take care of for this to be ready for review.

dgreif:
Sorry for the delay @djm181!  I'm curious what the use case is for this change.  Are you simply starting/stopping a lot as you develop another project?  I'm wondering if there is a better way to fetch current sessions from Ring so that we don't have to store the session elsewhere.  Also curious why you added the `[extra_meta:string] : any`.  If there are other details that you are seeing, i'd prefer to add the specific types.

djm181:
So as I mentioned in my feature request #378 (which I do plan to send in a PR for), I want to be able to disconnect the `RingApi` instance and reconnect with different options - if this is happening frequently then one can run into the limit, but it's definitely exacerbated during development. I would be a little surprised if Ring allowed fetching of sessions because the response includes an authentication token (although it's unclear what this is for).

As for the extra meta, that was a lazy catch-all but you are right - we should just introduce the missing fields. I'll push an update with those now.

dgreif:
Sounds good. I am hoping to get some time for Ring this weekend so I may be able to assess the session situation a little closer. Users sometimes run into the limit in homebridge ring if they have a bad plugin continually crashing their homebridge server, so this is definitely something I would like to address if possible. Do you use any of the values from the session response in your project? What is it that you are implementing if you don’t mind me asking?

dgreif:
I did some testing this morning and it looks like the session only needs to be created if Ring doesn't recognize the `hardware_id`.  The rest client will now only create a new session if it receives a response indicating that a session wasn't found.  I also added `api.getProfile()` which allows you to get the same profile response that the session endpoint provided.  These changes are pushed to master, and I plan to do a release today or tomorrow after I get a few more changes in.  Thanks for encouraging this change, I think it will be very useful for all users!

dgreif:
Changes released in v`9.4.0`

PR #399, opened 2020-07-11
Use Thermostat accessory to choose Snooze duration

dxdc:
- An improved variant of https://github.com/dgreif/ring/pull/365 OR https://github.com/dgreif/ring/pull/370
- Allows to set snooze hours in terms of degrees
- Simple toggle on/off button
- Working, but needs additional testing

### TODO: 
- Temperature units need to be converted to a normal standard (ideally celsius). I have tried setting `TemperatureDisplayUnits` but it didn't work so far.

https://github.com/homebridge/HAP-NodeJS/blob/5cacbd400bc15785f372eefb83bc40a5e21a439a/src/lib/gen/HomeKit.ts#L2953-L2975

- Might benefit from having the target temperature decoupled from the snooze duration.
- Config option to turn on/off as switch

### DIFF
--- a/homebridge/chime.ts
+++ b/homebridge/chime.ts
@@ -5,14 +5,32 @@ import { Logging, PlatformAccessory } from 'homebridge'
 import { BaseDataAccessory } from './base-data-accessory'
 import { logInfo } from '../api/util'
 
-const secondsFor1Hour = 60 * 60
-const elapsedTime = (elapsed: number) => {
+const secondsFor1Hour = 60 * 60,
+  elapsedTime = (elapsed: number) => {
     elapsed *= secondsFor1Hour
     const hours = Math.floor(elapsed / secondsFor1Hour),
-        minutes = Math.floor(elapsed / 60) % 60,
-        seconds = Math.round(elapsed % 60)
-    return `${hours}:${minutes.toString().padStart(2, '0')}:${seconds.toString().padStart(2, '0')}`
-}
+      minutes = Math.floor(elapsed / 60) % 60,
+      seconds = Math.round(elapsed % 60)
+    return `${hours}:${minutes
+      .toString()
+      .padStart(2, '0')}:${seconds.toString().padStart(2, '0')}`
+  }
+
+// Temperature conversion for C to F
+// Set useF based on your default HomeKit display
+//
+// Since HomeKit stores everything in Celsius, this results in some (minor) differences
+// in terms of snooze times. For example, 15:06:00 instead of 15:00:00 snooze. This could be
+// addressed differently if more accuracy is needed.
+//
+// Lastly, there are some oddities at the boundaries, so some numbers have been chosen manually
+// I.e., -17.77 or 25.5 for optimal behavior
+
+const roundSingle = (num: number) => Math.round((num + Number.EPSILON) * 10) / 10,
+  CtoF = (temp: number) => roundSingle(1.8 * temp + 32),
+  FtoC = (temp: number) =>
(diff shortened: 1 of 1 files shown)


dgreif:
Could get weird if you consider Celsius can farenheit... I’m fine with just doing the valve.

dxdc:
> Could get weird if you consider Celsius can farenheit... I’m fine with just doing the valve.

Yeah, that's what I meant by:
> Temperature units need to be converted to a normal standard (ideally celsius). I have tried setting TemperatureDisplayUnits but it didn't work so far.

It seems there should be a way to lock it. Without that, it is weird, but if there is a way to convert it fully to show Celsius only then it would be perfect.

dxdc:
@dgreif another option would be in the config, user could choose C or F as their default HK display units. Then, we can simply multiply all the units to convert.

Not sure if there is a cleaner solution? I do like the simplicity of this option though.

dgreif:
I think the valve works nicely because it has the time remaining built in. Every solution has pluses and minuses 😕

dxdc:
@dgreif give this latest version a try, if you would. I hard-coded `useF` but I think the functionality works really well actually. Curious on your take.

Something also off with the `elapsedTime` function... I saw this.
```
1|homebridge  | [7/11/2020, 3:53:32 PM] [Ring] Chime snoozed for 4:05:60
```

dxdc:
> I think the valve works nicely because it has the time remaining built in

Agree with this, but it's set to the temperature now. So, if 24 hours remaining, 24 degrees, etc.

dgreif:
@dxdc I'm still stuck on the fact that it can get messed up with F vs C.  I think we should take a step back and talk about why this is needed again. The current switch allows you to create automations that will turn the switch on and then turn it off after X minutes, which seems like it should cover most use cases in my mind.  For more complex automations, I assume you could use Node Red to handle the duration aspect.  What's the scenario in which the valve or thermostat is really beneficial?

dgreif:
@dxdc thoughts on closing this since we never made a decision on the best way to implement?

dxdc:
@dgreif Sorry for slacking on this one, I've been really busy the past few months. I experimented with it some more, but HK still lacks a great way to set a timer and everything else I've tried seems somewhat clunky and non-intuitive.

So, I think it's fine to close the issue for now.

I have implemented a variant of my idea inside NodeRED, which essentially listens for motion alerts, and then, if >X come in a fixed time window, it turns on the snooze and keeps it on if motion alerts continue at that rate -- and it turns them off afterward. It may still need some fine tuning, but it is workable.

One thing that could improve my set up is having a trigger that fires only when the Chime is played with a specific alert. I tried the "Speaker" sensor, but that didn't seem to fire when motion happened. Do you know if such a sensor exists? I would be fine with having a Motion sensor inside the Chime to detect when the Motion is played.

This all being said, I may be moving to more of a speech-to-text solution. E.g. "Motion at the Front", via Sonos. It may be more intuitive than these dumb chimes that are only good for a ring or motion MP3 :)

dgreif:
@dxdc no worries about being busy, I've been in the same boat 😄 .  I don't think there is any way to detect which sounds are being played on the Chime, only to tell it _to play a sound_, so I'm not sure there is a better solution on that front.  I haven't had any more complaints about this specific snooze duration issue, so I think we are good to close this.  I may finally install Node Red sometime soon as I also need to start setting up some more advanced automations and just can't quite do it all with HomeKit.  Good luck with the Sonos setup! And thanks for all your hard work looking into this and other issues on the repo 🙇

PR #474, opened 2020-10-13
Enable faster start of stream

djm181:
ffmpeg spends about 3 seconds analyzing the RTP stream to determine the details of the streamed video. This change instructs the RTP decoder to use a fixed frame size, and ffmpeg to spend nearly no time analyzing the incoming stream. If the native frame size of the camera doesn't match this then it has no adverse effects that I can observe. I find this allows streaming to start in less than 1 second.

cc @dgreif

### DIFF
--- a/api/sip-session.ts
+++ b/api/sip-session.ts
@@ -138,6 +138,8 @@ export class SipSession extends Subscribed {
         'pipe,udp,rtp,file,crypto',
         '-f',
         'sdp',
+        '-probesize', '32',
+        '-analyzeduration', '1000',
         ...(ffmpegOptions.input || []),
         '-i',
         'pipe:',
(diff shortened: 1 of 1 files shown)


dgreif:
Looks good at first glance, but I'd like to do some testing with it before I merge. Probably Wednesday or Thursday. Thanks a bunch for adding this tho! I've always been surprised by how slow ffmpeg starts when video is included, and this definitely explains why!

dgreif:
@djm181 I pulled this in and tested it in a branch.  Unfortunately, it doesn't look like ring is consistent with the video stream size.  You had it set to `1920x1080`, but I have seen some streams at `1280x720` and others at `640x360`.  If we don't have the size right, you get a green buffer around the actual video that fills the empty space.  I think it's worth it to keep the analyze period for now, unless you can think of a better way to determine stream size (which I can't).  Really appreciate the contribution, even if it didn't make it in!

djm181:
@dgreif This is interesting - I've experimented with a stickup_cam_v4, and a lpd_v2 and both of these seem to have the right behavior when I put in different sizes (I've tried putting in a size of the wrong aspect ratio too, and that works ok).

Do you have some examples of camera types that added the green buffer? I have access to several cameras and may try to reproduce to see if I can find a fix.

dgreif:
@djm181 I was testing it using `npm run record-example` within this project

PR #481, opened 2020-10-20
Faster start for streams

djm181:
@dgreif I'm taking a second go at this. Turns out `-vcodec copy` was the culprit for the green box. Given that the way I have this set up can be a bit intrusive I've elected to make it option-controlled. How do you feel about this approach? This makes the biggest difference with HLS streams I've found, but this also resolved the green box on other recordings.


dgreif:
merged to a branch for testing.  Seeing `Could not find codec parameters for stream 1 (Video: h264, none, 1920x1080): unspecified pixel format` when the stream starts, which I don't recall seeing the first time testing it.  Ideally i'd like to get the pixel format set, but options for that don't seem to work.  any additional advice?

djm181:
Is that message actually interfering with your stream? I definitely see some warnings from ffmpeg but it seems to be able to recover the video stream from it. I assume you've tried the obvious of adding a `-pix_fmt yuv420p` to the input stream parameters? I can't find any options in the x264 library that would be promising beyond that.

dgreif:
Yep, tried that option and it errored immediately. It does still stream, but the warnings specifically say it could use a longer analyze time to avoid issues. I'll play around with it a little more when I get some time

PR #571, opened 2021-02-13
Show charging status for battery doorbells

Sunoo:
This adds support for showing whether battery powered doorbells are hardwired by using the charging status. I tested this by disconnecting my doorbells from the wiring, and it works as expected there.

I only have access to two different models of battery doorbell, but I don't see any reason why this should cause issues with any other device types.

### DIFF
--- a/api/ring-camera.ts
+++ b/api/ring-camera.ts
@@ -118,6 +118,7 @@ export class RingCamera extends Subscribed {
       this.batteryLevel !== null &&
       this.batteryLevel < 100 &&
       this.batteryLevel >= 0)
+  isCharging = this.initialData.external_connection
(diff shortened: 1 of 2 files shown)


dgreif:
@Sunoo this is released in `9.15.0`.  Thanks for the addition!

jchristianj:
Not sure what that means. Should it display "charging: yes" in the detailed information page in home.app? Then for me it is not working. Have a Doorbell Gen 2 (cocoa_doorbell) and while it is charged on USB it says "charging: no".

Another question about battery here: Isn't there a warning for battery low in homekit for the Ring doorbells when battery level drops below 20%?

PR #625, opened 2021-04-26
Add support for updating motion settings for cameras

tokreutz:
I'm not sure which device types has this API, but for my stickup camera this is the only way for me to update the motion_detection_enabled setting. 

Using the existing subscribeToMotionEvents does not have any effect on the motion detection for my stickup camera.


tokreutz:
I made it more generic with my latest push. I added more or less (I omitted some things that seemed very configuration specific) the full object from the response from /devices/v1/devices/device_id/settings for getDeviceSettings and the properties that looked like they could be changeable using PATCH into an Update class.

@dgreif  Please have a look and suggest improvements. I havent tested all of the properties in the Update class for example, should I remove the ones that are "untested", or just leave them there, and then if they are not supported, that can be removed later?

tokreutz:
@dgreif, can you share some love with this PR? :)

dgreif (approved):

dgreif:
Thanks @tokreutz.  I'll merge it into a branch for some testing/touch-up and should have it released soon

dgreif:
@tokreutz released in v`9.18.0`.  In retrospect, I think you could have accomplished the same goal with motion settings by using `motion_detection_enabled: true` with `setSettings`, which is a different setting than what gets set by calling `subscribeToMotionEvents`.  Regardless, I don't mind supporting the new device settings endpoint if that's the direction Ring is headed, so thank you for the PR!

tokreutz:
Thats great @dgreif , happy to contribute with new functionality :)

Issue #624, opened 2021-04-23
Switch Simulation


MoTechnicalities:
# Feature Request

### Description
I got the new Outdoor Smart Plug but it only displays as a light.

### Describe the solution you'd like
I want to display it as a switch or as a fan.

dgreif:
The challenge is that Ring reports the plugins as a `switch.beams`, which simply indicates that it is a switch in their "Beams" aka Smart Lighting lineup.  I'm not sure there is a lot of value in making this configurable.  Do you have a specific reason you would like to change the type, other than to make it look right in the Home app?

MoTechnicalities:
I'm not sure why you would think there is little value in making the appearance configurable. The hole point of HomeBridge is to give us the flexibility ￼to make things work with HomeKit. HomeKit is by nature intended to be a visually appealing interface. Why should we be content and constrained then to show a switch as a 💡? Why not as anything I wish to run on that switch?

It is the "ability" and "flexibility" that makes the thing valuable. It is the essence of HomeBridge.

dgreif:
The problem is that configuring individual _accessories_ with a _platform_ plugin is a challenging.  If you have ideas on how that should be done, you are welcome to open a Pull Request and I'll take a look

MoTechnicalities:
I can recommend that you at least look into how the HomeBridge EWeLINK plugin does this as an example. The EWeLINK plugin is HOT!It allows Accessory Simulations. With it, I am even able to combine devices so that they perform the function of a HomeKit garage accessory. With it, a switch can easily be identified within the plugin as other types for HomeKit. I'm not a programmer, but having the ability do these types of configurations is what makes the plugin super valuable to me. Having the ability to use the Ring products within HomeKit is cool! But trust me, you can make it HOT!

MoTechnicalities:
I wanted to get back with you regarding a use case reason for allowing us to correctly configure the HomeKit simulation of the switch devices. I am currently connecting ventilation fans to my new outdoor switch, but with having HomeKit stuck on simulating the Ring Outdoor Switch as a 💡, instead of a fan, I am unable to simply tell Siri to turn off the lights in the Lab, as this will unintentionally cut off the fans as well. Such things don't happen in other rooms where I have fans connected to other brand smart plugs that are set to simulate fans in HomeKit.

Issue #322, opened 2020-05-12
Device categories occasionally revert (“Fans” → “Lights”)


smockle:
# Bug Report

### Describe the Bug
The Ring app for iOS allows you to specify a category for certain devices:
<img alt="ring-app-device-category" src="https://user-images.githubusercontent.com/3104489/81722137-f6f78d00-944e-11ea-8018-db436bea8d73.jpeg" width="407px">

Occasionally, after restarting `homebridge-ring`, I see messages like this in my logs:
```
Pruning unused service 00000040-0000-1000-8000-0026BB765291 Living Room Ceiling Fan from Living Room Ceiling Fan
```

This message indicates that my “Living Room Ceiling Fan” fan device was removed from the Home app. In its place, `homebridge-ring` has created a new light device. When I check “Living Room Ceiling Fan”’s category in the Ring app, I see that it has been reset from “Fans” back to “Lights” (the default).

Is it possible `homebridge-ring` is inadvertently overriding device categories on startup?

### To Reproduce
1. Add a device to the Ring app (for testing purposes, this can be a light device like the [GE Z-Wave Plus Dimmer Switch](https://byjasco.com/products/ge-z-wave-plus-wall-smart-dimmer) or an actual fan control like the [GE Z-Wave Plus Fan Control](https://byjasco.com/products/ge-z-wave-wall-smart-fan-control)).
2. In the Ring app, set the device’s “category” to “Fans“.
3. Restart `homebridge-ring`. Notice that the device is successfully added to HomeKit as a fan.
4. Wait.
5. Restart `homebridge-ring`. It’s possible your device now shows up as a light in the Home and Ring apps.

This describes the circumstances in which I experience the issue. Unfortunately, I don’t have a more specific repro step 4. I’m not sure how much time must elapse, or whether time is incidental to the actual cause.

### Expected behavior
I expect devices I configure as “Fans” to remain as “Fans” and not revert to “Lights”.

### Screenshots/Logs
To workaround this issue, I set the device category back to “Fans” in the Ring app, then restart `homebridge-ring`. Then, messages like this appear in my logs, showing that the errant light device is removed (a fan device appears in its place):
```
Pruning unused service 00000043-0000-1000-8000-0026BB765291 Living Room Ceiling Fan from Living Room Ceiling Fan
```

### Additional context
[...]

PR #631, opened 2021-05-02
feat: Support configuration to override device categories

smockle:
A possible solution to https://github.com/dgreif/ring/issues/624 and https://github.com/dgreif/ring/issues/322.

This PR adds support for a `categoryIdOverrides` configuration option, a map between device UUIDs and [category ids](https://github.com/dgreif/ring/blob/154f8fa817b7287fae28ced3c976c0de2fc31eae/api/ring-types.ts#L44-L62).

#### Warnings
⚠️ There are currently no guardrails—devices can be associated with unrecognized or invalid categories. For example, there’s nothing stopping someone from setting the `categoryId` of their light to `11` (thermostat), though the plugin wouldn’t facilitate setting that light to 5ºC, because accessory class is primarily [determined with `deviceType`](https://github.com/dgreif/ring/blob/154f8fa817b7287fae28ced3c976c0de2fc31eae/homebridge/ring-platform.ts#L125-L126), not `categoryId`.

⚠️ I have not run this locally yet (i.e. this code is untested), because I’m not convinced this PR is worth merging (ref the “Alternative”s below)—mostly opening it for discussion.

#### Alternative — No Manual Configuration
From @dgreif in https://github.com/dgreif/ring/issues/624#issuecomment-826484908:
> Ring reports the plugins as a `switch.beams`, which simply indicates that it is a switch in their "Beams" aka Smart Lighting lineup.

I don’t have any of these devices—does anyone know if Ring reports category ids for them? If so, we could add a `categoryId`-based conditional within the [`RingDeviceType.Beams*` case](https://github.com/dgreif/ring/blob/154f8fa817b7287fae28ced3c976c0de2fc31eae/homebridge/ring-platform.ts#L103-L108) of `getAccessoryClass`, like the ternaries in [`RingDeviceType.Switch`](https://github.com/dgreif/ring/blob/154f8fa817b7287fae28ced3c976c0de2fc31eae/homebridge/ring-platform.ts#L116-L120) and [`RingDeviceType.MultilevelSwitch`](https://github.com/dgreif/ring/blob/154f8fa817b7287fae28ced3c976c0de2fc31eae/homebridge/ring-platform.ts#L109-L113), and avoid the need for any manual configuration (i.e. I would recommend _not_ merging this PR if that’s the case).

#### Alternative — Group Configuration By Category
An alternative configuration schema I considered (i.e. besides a top-level `categoryIdOverrides` key) would require breaking plugin configuration changes, but may simplify configuration overall: Group related configuration options by category (example below):

```js
// In other words, this:
{
 "hideCameraMotionSensor": true,
 "sendCameraMotionNotificationsToTv": true,
 "hideCameraSirenSwitch": true,
 "cameraStatusPollingSeconds": 20,
 "cameraDingsPollingSeconds": 2,
 "categoryIdOverrides": { "f259b31c-826c-4b42-92f5-699927cbe032": 12 } // 12 is the category id for cameras
}

// —would be replaced by this:
{
  "camera": {
    "hideMotionSensor": true,
    "sendMotionNotificationsToTv": true,
    "hideSirenSwitch": true,
    "statusPollingSeconds": 20,
    "dingsPollingSeconds": 2,
    "overrides": ["f259b31c-826c-4b42-92f5-699927cbe032"]
  }
}
```

### DIFF
--- a/homebridge/README.md
+++ b/homebridge/README.md
@@ -68,7 +68,8 @@ Only include an optional parameter if you actually need it.  Default behavior wi
   "cameraStatusPollingSeconds": 20,
   "cameraDingsPollingSeconds": 2,
   "locationModePollingSeconds": 20,
-  "locationIds": ["488e4800-fcde-4493-969b-d1a06f683102", "4bbed7a7-06df-4f18-b3af-291c89854d60"]
+  "locationIds": ["488e4800-fcde-4493-969b-d1a06f683102", "4bbed7a7-06df-4f18-b3af-291c89854d60"],
+  "categoryIdOverrides": { "f259b31c-826c-4b42-92f5-699927cbe032": 2 }
 }
 ```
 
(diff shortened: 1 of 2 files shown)


MoTechnicalities:
Any progress in getting this out to us.

dgreif:
Thanks for the idea @smockle!  I haven't seen a huge need for this, so going to close for now

Issue #702, opened 2021-07-27
Updated to latest version and doesn't work 9.18.5 and get 404 error connecting to API


alangsam1:
# Bug Report

### Describe the Bug
 updated Ring Plugin to latest 9.18.5. and rebooted.Immediately got errors. Uninstalled and reinstalled on a separate RING Bridge and still doesn't work.

### To Reproduce
 - Start Homebridge
 - Attempt to start/ All other plugins load.
 then Ring doesn't open 

 - 7/27/2021, 2:49:47 PMRing BridgeRingERRORRequest to https://app.ring.com/rhq/v1/devices/v1/locations failed with status 404. Response body: 
7/27/2021, 2:49:47 PMRing BridgeRingERRORError connecting to API
7/27/2021, 2:49:47 PMRing BridgeRingERRORHTTPError: Response code 404 (Not Found)

### Expected behavior

 expected to see my camera images in the homekit addable to stream etc. 

### Screenshots/Logs
 If applicable, add screenshots and log output to help explain your problem.

### Additional context
 Add any other context about the problem here.

### Homebridge Ring Config
Post homebridge-ring platform config  _without sensitive information_
```

HIGHLIGHT THIS TEXT AND POST CONFIG HERE
DO NOT INCLUDE YOUR REFRESH TOKEN OR EMAIL/PASSWORD (replace them with xxx)

```

### Environment

 - OS: [e.g. Raspbian]
 - Node.js: [e.g. 12.14.1] 
 - NPM: [e.g 6.13.4] 
 - homebridge-ring: [e.g 6.0.3]
 - homebridge: [e.g. 0.4.50 or N/A]
 - hoobs: [e.g. 0.4.50 or N/A7/27/2021, 2:33:38 PMHOOBSRingConfiguring cached accessory f2878488-b3bd-4d62-8dec-09ef10b43932 Front Door
7/27/2021, 2:33:38 PMHOOBSRingConfiguring cached accessory b74668e3-8a43-41cc-8128-0d4943a6a258 Oceanside
7/27/2021, 2:33:38 PMHOOBSRingConfiguring cached accessory 9a817284-b4a5-4328-87b7-f50e4a4b5ca0 North Side 
7/27/2021, 2:33:38 PMHOOBSRingConfiguring cached accessory 4ba2ba7b-cb9e-4fc3-8d29-53a2e306d89c Side
[...]

Issue #699, opened 2021-07-27
Get locations fails with HTTP 404


wgbartley:
# Bug Report

### Describe the Bug
Request to https://app.ring.com/rhq/v1/devices/v1/locations failed with status 404.

### To Reproduce
- Start/restart Homebridge
- View logs

### Expected behavior
Expected behavior would be no error reported

### Screenshots/Logs
```
[7/27/2021, 1:22:37 PM] [Ring] Request to https://app.ring.com/rhq/v1/devices/v1/locations failed with status 404. Response body: 
[7/27/2021, 1:22:37 PM] [Ring] Error connecting to API
[7/27/2021, 1:22:37 PM] [Ring] HTTPError: Response code 404 (Not Found)
    at Request.<anonymous> (/homebridge/node_modules/homebridge-ring/node_modules/got/dist/source/as-promise/index.js:117:42)
    at processTicksAndRejections (internal/process/task_queues.js:95:5) {
  code: undefined,
  timings: {
    start: 1627406557813,
    socket: 1627406557957,
    lookup: undefined,
    connect: undefined,
    secureConnect: undefined,
    upload: 1627406557957,
    response: 1627406557985,
    end: 1627406557985,
    error: undefined,
    abort: undefined,
    phases: {
      wait: 144,
      dns: undefined,
      tcp: undefined,
      tls: undefined,
      request: NaN,
      firstByte: 28,
      download: 0,
      total: 172
    }
  }
}
```

### Additional context
I am also using ring-client-api in a separate script.  Ring.getCameras() also returns a 404.

### Homebridge Ring Config
```
    "platforms": [
        {
            "refreshToken": "xxx",
            "platform": "Ring"
        }
    ],
```

### Environment
 - OS: Running in oznu/homebridge docker container with Ubuntu 20.04.2 LTS host OS
 - Node.js: 14.17.1
 - NPM: 6.14.13
 - homebridge-ring: 9.18.5
 - homebridge: 1.3.4
 - hoobs: N/A

stubbz00:
I'm getting this error as well.  I think it is on the Ring side as they are currently having outages for Video on Demand and Live Calls (Doorbell Dings & Motions).

wgbartley:
You would think they would return something other than a 404 if there were server-side issues (like a 500-level error).

I'll keep an eye on https://status.ring.com to see when it clears up.  Hopefully the errors are related!

x5nder:
[...]

PR #708, opened 2021-07-27
I found the locations API

mkellsy:
With a bit of trial and error I found the locations API

It looks like you can fetch the raw locations with this URL
https://api.ring.com/devices/v1/locations

This is in response to issues #706 #702 #699


dgreif:
Thanks @mkellsy!  I wrote this a slightly different way in d3b26e823978aefde2c8500244e057d9e1f8bcff, but I appreciate you tracking down the url!

PR #791, opened 2021-10-12
Prevent login attempt for null refresh tokens

mkellsy:
I made 2 simple changes;

1. Added an if statement that only runs connectToApi() if the refreshToken is not false.
2. Changed Promise.all to Promise.allSettled. Little known function about Promise.all is it is only called if everything is resolved. If anything fails, it will hang forever. Changing this to Promise.allSettled will run even if some promises fail.

I think this is what is causing the network on raspberry pis to hang. I have tested this on armhf, arm64 and amd64 architectures.


markaedmondson:
Reinstalled.  Still no login prompt.  I think I mentioned it in the original ticket, but this issue is happening across all plugins, not just Ring.

PR #851, opened 2022-01-05
clear timeouts when RingApi.disconnect() is called

### LINKED ISSUE #850: Node Does Not Exit For One-Off Scripts
# Bug Report

### Describe the Bug
When using this library to get information about Ring devices, the script does not exit immediately once it is completed. When used with a serverless function (like AWS Lambda), the request continues to run until it times out.

### To Reproduce
```js
# test_ring.js
const getenv =  require("getenv");
const dotenv = require("dotenv")
const { RingApi } = require('ring-client-api');

async function run() {
  dotenv.config();
  const ringApi = new RingApi({
    refreshToken: getenv('RING_TOKEN'),
  });
  const locations = await ringApi.getLocations();
  const location = locations[0];
  const devices = await location.getDevices();
  ringApi.disconnect();
  console.log(devices.length);
  console.log('done');
}

run()
```

"done" is printed pretty quickly but node continues to run until you run "ctrl-c".
```sh
$ time node test_ring.js
61
done
^C

real	0m40.989s
user	0m0.644s
sys	0m0.195s
```

### Expected behavior
Script exits as soon as "done" is printed.

### Additional context
Using `wtfnode`, I was able to pinpoint that the script is not exiting due to the [`setTimeout` in rest-client.ts](https://github.com/dgreif/ring/blob/b14442f1f64b9922d84f7357f7dde8475d6bd5c6/api/rest-client.ts#L118-L122).

I can open up a PR but wanted to see what you think the approach should be. A constructor option to prevent the settimeout from being set, perhaps?

### Environment

 - OS: MacOS 11.6
 - Node.js: 12.20.2
 - NPM: 6.14.11
 - ring-client-api: 9.23.0

timtrinidad:
Fixes #850 by clearing timeouts set to refresh auth token when RingApi.disconnect() is called.

### DIFF
--- a/api/api.ts
+++ b/api/api.ts
@@ -310,5 +310,7 @@ export class RingApi extends Subscribed {
       .catch((e) => {
         logError(e)
       })
+
+    this.restClient.clearTimeouts()
   }
 }
--- a/api/rest-client.ts
+++ b/api/rest-client.ts
@@ -104,6 +104,7 @@ export class RingRestClient {
   public refreshToken = ('refreshToken' in this.authOptions ? this.authOptions.refreshToken : undefined)
   private hardwareIdPromise = getHardwareId(this.authOptions.systemId)
   private _authPromise: Promise<AuthTokenResponse> | undefined
+  private timeouts: ReturnType<typeof setTimeout>[] = []
   private clearPreviousAuth() {
     this._authPromise = undefined
   }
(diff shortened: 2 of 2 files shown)


dgreif:
Were you able to confirm that this allows the client to exit on a call to `disconnect`?

timtrinidad:
It worked when running the script locally. I'm trying now after deploying to AWS Lambda but it's still timing out. I'll let you know what I find so we can keep this PR open for now.

timtrinidad:
I'm having a lot of trouble getting serverless-webpack to honor my local changes with `npm link`, so I can't test on AWS Lambda. Testing locally, however, I was able to get the script to exit as expected.

timtrinidad:
OK - I was able to figure out how to deploy to lambda with my local changes and it exits as expected.

PR #859, opened 2022-01-18
Decouple ffmpeg input argument generation from ffmpeg execution

koush:
This change improves the a/v pipeline within my use case. It should be a non-breaking change for everyone else.

This builds on top of prior pull request: https://github.com/dgreif/ring/pull/858

### DIFF
--- a/api/sip-session.ts
+++ b/api/sip-session.ts
@@ -158,38 +158,24 @@ export class SipSession extends Subscribed {
     }
   }
 
-  private startTranscoder(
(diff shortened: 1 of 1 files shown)


dgreif (changes requested):
@koush I'm good with these changes.  I think you are probably the only one that will use the new function, but I'm good with adding it.  Would you mind getting this rebased and fixing the lint issues?

koush:
Rebased and ran lint which seems to have fixed the formatting issues here.

dgreif (approved):

PR #891, opened 2022-02-18
split out live call negotiation

koush:


api/peer-connection.ts
```diff
@@ -46,7 +46,7 @@ export class PeerConnection {
           ],
         },
       })),
-      audioTransceiver = pc.addTransceiver(this.returnAudioTrack, {
+      audioTransceiver = pc.addTransceiver('audio', {
```
dgreif, line 49: hmm was this change needed?

koush: hm no I had this for testing my weird connection drops and forgot to take it out

koush:
alright the latest is merged and changes minimized. didn't realize you had already removed sip from the library.

koush:
was testing this endpoint and found that it is returning a sdp that seems to be specifically for clients that can handle H264 high. Presumably this is because they know/think the client is android. in any casem it fails on safari and chromecast, etc. I'll readd the original endpoint I had before as a fallback option for browsers.

koush:
Confirmed that reintroducing the browser endpoints resolved my Safari/Chromecast issues. https://github.com/koush/scrypted/commit/9850eebdda6a9b909acb5f55916b92e56b551806

dgreif:
I've added the ability to pass in a custom `createPeerConnection` option in the `beta` branch, so this should no longer be needed. Thanks for continuing to work with me on this api!

PR #899, opened 2022-02-28
allow option to disable retry mechanism on no response

koush:
I need a way to be able to create requests with the ring api, with no retries. the current rest client will always retry on no response, which happens on snapshots. I need to manage snapshot recovery myself, and that involves knowing when the request fails, and without any retries at all.

### DIFF
--- a/api/rest-client.ts
+++ b/api/rest-client.ts
@@ -49,7 +49,7 @@ export interface ExtendedResponse {
 }
 
 async function requestWithRetry<T>(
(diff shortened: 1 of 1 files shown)


koush:
This code also seemingly gets into a retry loop that does not ever seem to exit? I think it is currently not manifested in the API because the snapshot request is debounced against a single promise.

dgreif (approved):

dgreif (approved):

PR #946, opened 2022-05-22
Adding ffmpegPath

donavanbecker:
Adding ffmpegPath so that when someone has it in their config, it doesn't delete out of their config on a new save of the config in the UI. If you don't have this then the config gets removed on a save.

This will not display in the UI though.


dgreif (approved):

donavanbecker:
Thanks for merging!

PR #952, opened 2022-05-25
add esnext build outputs, make werift a runtime dependency by passing it as an RingApi option

koush:
just looking for comments atm.
main "breaking" change is passing in the werift connection factory into the constructor. not ideal, but it's the only way I was able to remove the dependency during bundle.


api/rest-client.ts
```diff
@@ -158,6 +154,11 @@ export class RingRestClient {
         ? this.authOptions.refreshToken
         : undefined
     this.hardwareIdPromise = getHardwareId(this.authOptions.systemId)
+    this.baseSessionMetadata = {
+      api_version: apiVersion,
+      device_model:
+        this.authOptions.controlCenterDisplayName ?? 'ring-client-api',
+    }
```
dgreif, line 161: Just curious, why was this moved?

tsconfig.esnext.json
```diff
@@ -0,0 +1,24 @@
+{
+  "compilerOptions": {
+    "target":"esnext",
+    "lib": [
+      "es5",
+      "es2015",
+      "es2017",
+      "dom"
+    ],
+    "module": "commonjs",
```
dgreif, line 10: This is a little confusing.  Do you intend to have ESM output here?  If so, `module` should be `ES2020` or higher I believe.

api/rest-client.ts
koush: initialization order issue in the constructor. it isn't set when it's referenced.

dgreif:
@koush I've redone the repo as a true monorepo. I'm not sure how many of your changes from this PR as still applicable. I'd love to keep scrypted on the main version of `ring-client-api` as I know you have made some good tweaks over time. Feel free to open a new PR if you are interested in removing your fork.

koush:
I’m pretty far diverged now with past pull requests and additional changes, and don’t really have the time to merge back now. Too much work.

PR #1150, opened 2023-02-06
Enable retransmission of lost RTP packets via RTX

tsightler:
Hi @dgreif ,

For months I've been plagued by random artificing when viewing streams with ring-client-api and I finally had time to dig into it.  For a while I thought this was some problem with werift itself, perhaps just poor performance overall, but after adding some debugging it was clear that, for my case UDP packets were occassionally lost before they reached even my firewall.

What didn't make sense was that videos viewed with the Ring app or Ring web dashboard on the same device/network never experienced this issue so I took some packet traces and really started digging into it and finally I noticed that both the Ring web dahsboard and app use RTX to support retransmission of lost packets.  You can see the same occassional packet loss in the streams viewing in Google browser, but any loss followed quickly by RTCP feedback telling the upstream that packets were lost, and then, within 100ms or so, a retransmit on the channel via RTX payload type.

Taking the same trace from the same Ring media server by using ring-client-api with werift showed that a packet would be lost, and werift would send RTCP feedback just the same, but no retransmitted packet was ever seen.  On further digging I noticed that the SDP offer received via the websocket included RTX, but the answer created by werift did not, thus there was no way for packets to retransmit even though feedback was being sent.  Adding this small patch allows werift to produce and SDP answer that includes RTX payload type and indeed, the behavior then matches what is seen when using a browser.  This dramatically decreases artifacts in my streams.  Please consider including in a future version.

### DIFF
--- a/packages/ring-client-api/streaming/peer-connection.ts
+++ b/packages/ring-client-api/streaming/peer-connection.ts
@@ -82,6 +82,10 @@ export class WeriftPeerConnection
               parameters:
                 'packetization-mode=1;profile-level-id=640029;level-asymmetry-allowed=1',
             }),
+            new RTCRtpCodecParameters({
+              mimeType: "video/rtx",
+              clockRate: 90000,
+            }),
           ],
         },
         iceServers: ringIceServers.map((server) => ({ urls: server })),
--- a/packages/ring-client-api/streaming/peer-connection.ts
+++ b/packages/ring-client-api/streaming/peer-connection.ts
@@ -83,7 +83,7 @@ export class WeriftPeerConnection
                 'packetization-mode=1;profile-level-id=640029;level-asymmetry-allowed=1',
             }),
             new RTCRtpCodecParameters({
-              mimeType: "video/rtx",
+              mimeType: 'video/rtx',
               clockRate: 90000,
             }),
           ],


dgreif (approved):
I tried it out and don't notice any differences on my cameras, but I also don't notice many issues with my streams. I trust you've tested it thoroughly, and I really appreciate you digging so deep into this issue! I'll try to get this merged/deployed at some point this week

tsightler:
Yeah, if you don't have many artifacts, I wouldn't expect much difference.  It seems like the problem showed up a few months ago, after my ISP (a small, regional FTTH provider) added some additional capacity upstream.  My route in/out of the AWS network changed pretty significantly and I started seeing the artifacts like the one below fairly regularly (say, a dozen times in 60 seconds):
artifact

I've tested it quite a bit and, at worst, I don't think it can have any negative impact, it simply lets the local werift WebRTC instance use RTX if it is offered, which, in my testing with the various Ring WebRTC endpoints, it always is.  Also, it's extremely standard to use RTX with any video stream, exactly because UDP is lossy, there are many examples all over of its use.

I still think the werift RTCP feedback code could use some improvement, it's not as efficient/reliable as the Chrome implementation, and it seems to go a little wild requesting retransmits too quickly, IMO, but I'm still digging into this more, slowly wrapping my brain around the various code paths in werift and how all these protocols fit together.

dgreif (approved):

packages/ring-client-api/streaming/peer-connection.ts
```diff
@@ -82,6 +82,10 @@ export class WeriftPeerConnection
               parameters:
                 'packetization-mode=1;profile-level-id=640029;level-asymmetry-allowed=1',
             }),
+            new RTCRtpCodecParameters({
+              mimeType: "video/rtx",
```
dgreif, line 86:
```suggestion
              mimeType: 'video/rtx',
```

PR #1221, opened 2023-05-22
Return information for both batteries

desbo:
For devices that support two batteries, the current logic to get the battery level returns the minimum of the two batteries.

I need to check the level of both batteries separately and think it would be preferable not to discard the greater of the two levels. I have updated the code to return an object with `primary` and `secondary` fields to show the level of each battery for devices with two batteries. In single-battery devices, only the `primary` field will have a value.

### DIFF
--- a/packages/ring-client-api/ring-camera.ts
+++ b/packages/ring-client-api/ring-camera.ts
@@ -65,25 +65,26 @@ function getEndOfToday() {
   return getStartOfToday() + fullDayMs - 1
 }
 
+interface BatteryLevel {
+  primary: number | null,
+  secondary: number | null
+}
+
 export function getBatteryLevel(
   data: Pick<CameraData, 'battery_life' | 'battery_life_2'> & {
     health?: Partial<CameraData['health']>
   }
-) {
-  const levels = [
-      parseBatteryLife(data.battery_life),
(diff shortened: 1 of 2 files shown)


dgreif:
Hey @desbo, thanks for taking the time to open a PR! Unfortunately this would be a breaking change to the api, which is something I'd like to avoid. While I like the idea of having this data neatly exposed by the `RingCamera` class, I'm curious if you could just implement the same logic in your use case by looking at `camera.data.battery_life` and `camera.data.battery_life_2`?

desbo:
@dgreif thanks for the suggestion. I've found that doesn't compile because the type of `camera.data` is `CameraData | OnvifCameraData` and `OnvifCameraData` doesn't have the battery life fields. My TS knowledge is a little patchy though, so I might be missing something — any thoughts?

dgreif:
Ahh, makes sense. You should be able to force typescript to use the desired types with something like `const data = camera.data as CameraData` and then `data.battery_life` should resolve properly

desbo:
@dgreif thanks! I also found checking for `"battery_life" in camera.data` allowed me to access the fields. I'll close this as I have what I need. Thanks a lot for the library!

PR #1278, opened 2023-08-01
Update and consolidate streaming API

tsightler:
This PR removes the old app based streaming methods and consolidates all streaming to use the new methods used by the Ring web based UI.

This new method appears to work identically regardless of legacy cameras using H.264/AVC or newer cameras using H.265/HEVC or for Ring Edge devices.  For H.265/HEVC cameras this API appears to simply transcode the stream back to H.264/AVC on the fly when the client sends an offer that doesn't include H.265.  I think the reason the browser based method supports this, while the app based one does not, is that many browsers do not yet support H.265 via WebRTC although they will all mostly play back H.265 so playback of recoded events does work.

The API itself is nearly identical to that previously implemented in ring-edge-connection.ts with the exception of requiring a specific API call to acquire a ticket for authentication, and the websocket URL containing all of the connect options instead of the in the request headers, but it is otherwise the same endpoint and the socket messages themselves appear identical.

Also had to add dialog_id to all sendMessage calls (perhaps should have just been implemented in sendMessage function now that I think about it).  The dialog_id is, as far as I can tell, just a random UUID used for a given streaming session, but without it the signaling socket would close the connection immediately on any sent message.

The actual code changes have been pushed in ring-mqtt for over a week, with ~80% of users already upgraded due to automatic updates, and, so far, no reports of new issues, but I can't rule out possible problems.  The only thing I have not personally tested is Ring Edge, but at least two users reported that it worked.

This is my first attempt at pushing a patch into a Typescript project, so go easy on me @dgreif!


dgreif (approved):
This is great progress, and appears to work perfectly with both cloud and Ring Edge devices. Great work @tsightler! I'll merge this and do some additional cleanup on my end. Thank you!

packages/ring-client-api/ring-camera.ts
```diff
@@ -378,27 +377,16 @@ export class RingCamera extends Subscribed {
...
         method: 'POST',
-        url: this.doorbotUrl('live_call'),
+        url: 'https://app.ring.com/api/v1/clap/ticket/request/signalsocket',
       })
       .catch((e) => {
-        if (e.response?.statusCode === 403) {
-          const errorMessage = `Camera ${this.name} returned 403 when starting a live stream.  This usually indicates that live streaming is blocked by Modes settings.  Check your Ring app and verify that you are able to stream from this camera with the current Modes settings.`
-          logError(errorMessage)
-          throw new Error(errorMessage)
-        }
```
dgreif, line 396: Fun fact - I assumed we would still get these 403s for cameras that had live view disabled due to Modes. Looks like that's not the case! Just tried it and it let me stream from a camera which was supposed to be disabled. Ring app won't let me, but I could stream via ring-client-api 😂. Not sure if we should call this a bug or a feature, but thought I'd note it here 😄

tsightler: Yep, exactly why I removed this because, to my surprise, it just streamed anyway.  It didn't surprise me that a ticket would still be issued, but I was surprised the socket still accepted the livecall command.  I decided to just allow this in the ring-client-api PR since the Ring API technically allows it and leaves it up to the client side to enforce.

I'm not saying this is the right decision, just the one I made at the time.  I personally saw it as a feature, but I haven't decided whether to enforce or not in ring-mqtt itself.

PR #1307, opened 2023-09-06
Fix ffmpeg error when streaming over LTE/5G

tsightler:
Super small fix here.  I appears the tweaks to Opus encoding which disabled variable bitrate encoding caused ffmpeg to produce RTP packets which slight exceed the current maximum configured packet size of 188 (176 data bytes).

This size works over local networks because Homekit requests a packet time (frame size) of 20ms which produces packets of ~60-80 bytes.  However, over LTE/5G network Homekit requests packet time of 60ms, which produces packets of 3x that size, or 180-240 bytes, which is over the hardcoded 188 byte size.  This causes ffmpeg to exit unexpectedly due to RTP packet size too big error and thus streaming will not work over LTE/5G networks (I'm not convinced this is not a bug with ffmpeg, as I would have expected the packet to be fragmented, but I don't really understand the code path here).

However, simply increasing the packet size seems to prevent the ffmpeg process from crashing.  This probably will not fix all issues with LTE/5G streaming, but it will fix this critical issue which keeps streaming from working in most cases.

I'm not 100% sure why this value was set to begin with.  In my testing the 276 byte value seems to be sufficient, but I wonder if we need any setting here at all since packet size should largely be dictated by bitrate and frame size parameters which are already explicitly set.  @dgreif perhaps you have some history here you could share?

### DIFF
--- a/.changeset/poor-moles-cry.md
+++ b/.changeset/poor-moles-cry.md
@@ -1,0 +1,6 @@
+---
+'homebridge-ring': patch
+---
+
+ Update Opus processing to handle all possible bitrates and packet lengths which should fix audio issues over LTE/5G
+ Minor tweaks to AAC-ELD processing for slightly improved audio
--- a/packages/homebridge-ring/camera-source.ts
+++ b/packages/homebridge-ring/camera-source.ts
@@ -157,11 +157,11 @@ class StreamingSessionWrapper {
         audio: { port: audioPort },
       } = this.prepareStreamRequest,
       {
-        audio: { codec: audioCodec, sample_rate: audioSampleRate },
+        audio: { codec: audioCodec, sample_rate: audioSampleRate, packet_time: audioPacketTime },
       } = startStreamRequest,
       // Repacketize the audio stream after it's been transcoded
-      opusRepacketizer = new OpusRepacketizer(1),
-      audioIntervalScale = audioSampleRate / 8,
+      opusRepacketizer = new OpusRepacketizer(audioPacketTime / 20),
+      audioIntervalScale = (audioSampleRate / 8) * audioPacketTime / 20,
       audioSrtpSession = new SrtpSession(getSessionConfig(this.audioSrtp))
 
     let firstTimestamp: number,
@@ -173,7 +173,7 @@ class StreamingSessionWrapper {
       if (audioCodec === AudioStreamingCodecType.OPUS) {
         // borrowed from scrypted
         // Original source: https://github.com/koush/scrypted/blob/c13ba09889c3e0d9d3724cb7d49253c9d787fb97/plugins/homekit/src/types/camera/camera-streaming-srtp-sender.ts#L124-L143
-        rtp = opusRepacketizer.repacketize(RtpPacket.deSerialize(message))
+        rtp = opusRepacketizer.repacketize(rtp)
 
         if (!rtp) {
           return null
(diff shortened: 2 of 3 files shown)


dgreif (approved):
Thanks for the fix, and for digging into these really obscure bugs. The 188 packet size is something I played with a bit when I first set up HomeKit streaming, but haven't touched in a long time. I likely pulled the 188 from another streaming plugin at the time, and I remember trying larger/smaller values and going with this since it appeared to provide the smoothest stream. I think bumping it makes sense, so let's get this released. 

One small favor, could you run `npx changest` on this branch and add a changeset which explains what's in this PR? I'd consider this a "patch" change. Once that's added, we can squash/merge and release

tsightler:
@dgreif Sorry for the delay, but I've really been digging into the entire streaming pipeline to better understand it and, now that I understand it better, as well as the way Homekit expects audio, I think there is likely a more technically correct fix that should be made, I just need some additional time to complete the testing.

tsightler:
Hi @dgreif,

So I think this has reached the stage where it is ready to merge.  I know this patch looks a bit more scary, but it's actually not quite as bad as it seems.  Mostly it does the following:

**Updated Opus processing logic**
This puts the code in homebridge-ring on par with the current state in Scrypted, on which it appeared the original code was based.  There were a few funny things in the original code, like the fact that opusRepacketizer was always initialized to produce one packet per frame, which effectively rendered it a no-op.  We could just throw it away completely in that case.

The new code allocates the frame per packet based on requested frame size and also properly scales the RTP timestamp for sample rates other than 24Khz (at least Apple watch seems to request 8Khz).  This also seems to fix variable bitrate encoding, which is technically what the spec requires, so I've re-enabled that at well.

**Updated Opus return audio**
I finally tracked down the weird bug that was causing packets to be dropped and caused the super choppy return audio when Opus VBR mode was enabled (as it should be).  This turned out to be caused by an attempt to decrypt RTP payloads that were not encrypted.  I'm not really sure why that was ever in there, my gut says maybe in the early days you were trying to just copy the Opus packets received from Homekit directly to Ring and decrypt them there?  But, currently, return audio is always transcoded, even for Opus, so it's ffmpeg that decrypts the inbound audio from Homekit, and it outputs unencrypted RTP.  However, the returnAudioTranscodedSplitter still attempts to decrypt these packets, which fails at a high rate.  Honestly, I have no idea how it ever works, it seems like if packets are over a given size the decrypt function just silently returns the packet unchanged.

Anyway, I've removed the decrypt functions and the variables used to store the keys, since those aren't needed, and it seems to work with variable bitrate again.

**Minor tweaks to AAC-ELD encoding options**
Looking at the Unifi Protect homebridge plugin I saw some interesting options for AAC encoding and, after some research and trying them out, I have to say it seems to improve the audio quite a bit.  It's still not as good as Opus, but I think it's mostly due to the fact that Homekit is only using 16k sampling and 24k bitrate, which is below the ideal for AAC-ELD, but these options do seem to bring it closer.

Even with these tweaks the audio with Homekit is not as good as the Ring app, but I think this is mostly due to the lower sampling and bitrate.  When I compare ring-mqtt audio, which streams via RTSP, to the Ring app, it is indistinguishable (as it should be, as it's using ffmpeg with copy), but Homekit uses a sample rate of 24Khz instead of 48Khz, and a bitrate of no higher than 24Kbps vs native which is more like 64Kbps.  I don't know if Homekit supports 48Khz or higher bandwidths in newer specs, but this is something I need to research more.

**Increase maximum packet size**
This fixes the ffmpeg crash when doing LTE/5G due to the larger Opus packets.  376 seemed completely safe and had no measurable negative impact on AAC-ELD, so I went with it.  Just for reference, Scrypted uses 400, but appears to have fully deprecated AAC-ELD and is moving toward Opus 100% it seems.  

Regardless, I'm hopeful bringing this code up to par with current Scrypted will help with these issues.  It seems to work well for me, but there of course could still be problems I haven't uncovered.

dgreif (approved):
Gave these a try locally (at least the Opus side) and everything looks great! Thank you for all your hard work on this @tsightler, I think this will be a huge improvement 😁

Issue #1303, opened 2023-08-26
Live streaming works on Local Network but does not work on Remote Network - Ring Plugin 12.0.0


JOMA519:
### Streaming Issue

I'm hoping you can help me out, please!  I can't figure out this issue.  I have home bridge all setup on a clean install of Ventura on a Mac mini (wired connection).  I can't seem to get my ring doorbell to connect in home view or live view when I'm not on my local area network.

I get the following error in homebridge when I'm trying to access my live stream on my home app using cellular data:
[homebridge-ring] From Ring (Front Door): exited with code 1 and signal null.

Here are my config details:
- node.js version 20.5.1
- npm version 9.8.0
- ring plugin version 12.0.0
- homebridge version 1.6.1
- The ring plug-in is the only thing I'm using in homebridge.  All my other devices are HomeKit native.

My MacMini is set to not sleep.

I have reset my two Apple tv's and my two HomePods as some users have said to restart.  I've restarted my router.

Any help or suggestions would be greatly appreciated!  I have looked at past issues and discussion but there is no clear suggestions on how to fix.

### Proposed Solution

I'm guessing the only two options left to try:

- tweaking config.json files?
- possibly opening up ports on router but I have never had to do this before.

### Environment

Homebridge

### What operating system are you on?

Mac

### Honesty Time

- [X] I swear this is worth your time!

tsightler:
Other than what is already written in the Wiki page about this, there's not much else to suggest.  From the plugin perspective there's no difference between streaming locally vs streaming to a device on LTE as the plugin simply streams to the local home hub and it's up to the home hub to get that data to the client.

You should verify that whatever device you have designated as the Home Hub is definitely working and is on the same network with Home Bridge and that there are no ports being blocked by firewall rules or anything else and that you have no firewalls or VPNs or anything that would impact DNS or UDP traffic.

Note that using Homekit with the Ring plugin over LTE is very demanding on both the local network, due to all of the hops, and on the LTE network itself so it depends on having near perfect network connectivity through the entire path, which is quite complex.  In the end, this is all outside of the control of the plugin.

JOMA519:
Hi, really appreciate the response and getting back to me.  I have installed scrypted and the live video stream works perfectly on my 5G cellular connection.

For now, I have hidden the doorbell device if in Homebridge so it doesn’t show up on HomeKit.

I really would prefer not to have use scrypted as I Don’t need hksv features and having another service to maintain.  However this will work for now.

[...]

PR #1316, opened 2023-09-25
Update camera-source.ts to work in remote HomeKit connection

oliveira131:
This fix  `exited with code 1 and signal null` error 

When request.audio.packet_time is equal to 60 (remote on LTE/5G connection), ffmpeg fail and with -application voip or lowdelay mode.
In local requests network, the request.audio.packet_time is 20 and works fine.

I think this is the same error as we see in https://github.com/dgreif/ring/issues/1303

Maybe an "if" works too, but I don't see benefits (performance improvement) with this option.

### DIFF
--- a/packages/homebridge-ring/camera-source.ts
+++ b/packages/homebridge-ring/camera-source.ts
@@ -275,8 +275,8 @@ class StreamingSessionWrapper {
               'libopus',
               '-frame_duration',
               request.audio.packet_time,
-              '-application',
-              'lowdelay',
+              //'-application', //commented to work in mobile context. 
+              //'lowdelay',
             ]
           : [
               // AAC-eld specific


tsightler:
Can you please post the ffmpeg error that you see with 12.0.1?  It makes no sense that it would be the same as #1303 because that was caused by pkt_size=188 which was changed from 12.0.0 to 12.0.1 (as well as the entire Opus repacketizer process) and should have eliminated any packet size error from ffmpeg.  I'm not able to reproduce any such errors in 12.0.1 with lowdelay and that parameter is used in other similar projects also when no reports of issue.  I'm happy to consider it, but I'd like to understand specifically what error message you are seeing from ffmpeg.

tsightler:
I'm going to close this because this is being completely re-worked for the next version so that Opus does not need ffmpeg at all.

PR #1334, opened 2023-10-17
Update ring-types.ts

jarodwsams:
Add model info for "Video Doorbell 4" and "Spotlight Cam Plus" products.


dgreif (approved):
Thanks for getting these added! They sure are using some creative names now 😄

PR #1373, opened 2023-12-06
Motion detection switch

davidrthomas:
Add a motion detection switch characteristic to cameras. 
Allows individual camera motion detection to be enabled/disabled independently to the ring modes.

Use case: An automation to turn cameras on at night when alarm is disarmed.


davidrthomas:
Hi @dgreif any chance this could be considered in an upcoming release?

dgreif:
@davidrthomas sorry this sat for so long. I don't see a strong need for this feature so I'm going to close this for now.

PR #1386, opened 2024-01-09
ffmpeg: Expose stdout callback to user

keur:
Integrating in https://github.com/homebridge/camera-utils/pull/584

Would appreciate if we could do a new release once this is merged so I can use in my application.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1668,15 +1668,15 @@
       }
     },
     "node_modules/@homebridge/camera-utils": {
-      "version": "2.2.4",
-      "resolved": "https://registry.npmjs.org/@homebridge/camera-utils/-/camera-utils-2.2.4.tgz",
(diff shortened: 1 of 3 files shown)


keur:
cc @dgreif 

Don't have permission to request reviewers

dgreif:
@keur I'll go ahead and merge this, but I'm not sure we need to do a release just for this bump. Since we allow patch/minor updates for the `@homebridge/camera-utils` package, you should be able to install the latest version of that package and have ring automatically start using it.

keur:
@dgreif Thank you!

The way I need to `startTranscoding()` has changed with this new callback option. I understand if you don't want to do a release. I'm fine pointing the package at `git+https://github/dgreif/ring#main`. I am doing that right now to point at my fork. However, I am including it from a JS project. That means I need to go into the `node_modules` and manually build the `ring-client-api` package, then include the `lib` directory. My final include ends up looking like this: `const { RingApi } = require('ring/packages/ring-client-api/lib')` 

It gets a little ugly. Is there a better way to do that? I don't program JS regularly, so apologies if these are basic questions.

Issue #1312, opened 2023-09-20
No battery level reported for stickup_cam_v4


sjonez:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe The Bug

I'm not seeing any battery data for one of my cameras which is a stickup cam v4. Here's the discovery data:

 ```
[
	{
		"kind": "stickup_cam_v4",
		"description": "Garden",
		"is_sidewalk_gateway": false,
		"created_at": "2022-07-29T13:49:36Z",
		"deactivated_at": null,
		"firmware_version": "Up to Date",
		"owned": true,
		"stolen": false,
		"shared_at": null,
		"health": {
			"ac_power": 0,
			"battery_percentage_category": "unknown",
			"battery_save": false,
			"battery_voltage_category": "unknown",
			"connected": true,
			"device_type": "stickup_cam_v4",
			"external_connection": false,
			"firmware_version": "cam-12.0.38",
			"floodlight_on": false,
			"last_update_time": 1695194126,
			"network_connection_value": "wifi",
			"night_mode_on": false,
			"ota_status": "timeout",
			"rss_connected": true,
			"rssi": -70,
			"rssi_category": "okay",
			"second_battery_percentage": 61,
			"second_battery_percentage_category": "good",
			"second_battery_voltage_category": "good",
			"sidewalk_connection": false,
			"white_led_on": false,
			"firmware_version_status": "Up to Date",
			"packet_loss": 0,
			"packet_loss_category": "good",
			"hatch_open": false,
			"ext_power_state": 0,
			"second_battery_voltage": 3656,
			"active_schedule_uuid": "537bd0d37f6f45e9b35a02651fdcebfb",
			"bandwidth": 4800,
			"status_time": 55547722363647,
			"vod_enabled": true,
			"battery_error": false,
			"bandwidth_last_time": 1691548239,
			"firmware_avg_bitrate": "0",
			"stream_resolution": 0,
			"video_packets_total": "1423",
			"wifi_is_ring_network": false,
			"wifi_name": "Nacho Internet",
			"active_battery": 2
		},
		"subscribed": false,
		"subscribed_motions": true,
		"external_connection": false,
		"active_schedule_uuid": "537bd0d37f6f45e9b35a02651fdcebfb",
		"alerts": {
			"connection": "online",
			"ota_status": "timeout"
		},
		"motion_snooze": null,
		"battery_life": null,
		"settings": {
			"motion_snooze_preset_profile": "low",
[...]

PR #1422, opened 2024-05-26
Fix batteryLevel to check secondary battery

stanaka:
The current logic checks only the primary battery even each camera has a secondary battery.
This PR fixes this issue by checking the secondary battery in getBatteryLevel method.

Related issue: #1312


dgreif (approved):
Thanks for the updated logic! I don't have any cameras with two batteries so I haven't run into corner cases associated with the second battery

PR #1440, opened 2024-06-25
Don't send old FCM tokens when new token is generated during startup

tsightler:
This fix stops the code from pushing the old FCM token to Ring when a new token is generated during registration, for example, due to migration to FCMv1 APIs for registration.  The bug happens because the very last command in the push notification initialization sequence currently sends the FCM token to Ring any time  `credentials` is defined:
```
if (credentials) {
  onPushNotificationToken.next(credentials.fcm.token)
}
```
The problem with this is `credentials` is assigned prior to registration so it would still hold the old FCM token even if a new token was generated during the registration process.  Because of this, the sequence of events was something like this:

1) credentials = config.pnc (assuming credentials already existed)
2) Push notification would register using existing credentials
3) If registration updated the FCM token, `onCredentialsChanged()` was called which subsequently pushed the new generated FCM token to Ring
4) However, because `crendentials` was previously defined and still contained the old FCM token, the command noted above would also run immediately pushing the old FCM token back to Ring thus making it impossible to receive notifications with the new push registration

The new code only pushes the stored FCM token if it was unchanged during registration.

### DIFF
--- a/packages/ring-client-api/api.ts
+++ b/packages/ring-client-api/api.ts
@@ -233,7 +233,8 @@ export class RingApi extends Subscribed {
     intercoms: RingIntercom[],
   ) {
     const credentials =
-        this.restClient._internalOnly_pushNotificationCredentials?.config && this.restClient._internalOnly_pushNotificationCredentials,
+        this.restClient._internalOnly_pushNotificationCredentials?.config &&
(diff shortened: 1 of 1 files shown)


dgreif (approved):
Merging as-is and will fix up in the beta branch

PR #1507, opened 2024-10-14
Handle iceCandidate event with no candidate

tsightler:
This PR addresses compatibility with WeRIFT v.0.20.0 and later which modified onIceCandidate behavior to be the same as browser API.  Receiving an iceCandidate event with no candidate means all candidates are gathered.

This code just ignores evens with no candidate property, but we could also add an else and log ICE gathering complete for debug purposes if you think that is useful.


dgreif:
I've got a fix for this in the beta branch now. Thanks for the heads up and initial research!

PR #1516, opened 2024-10-28
Fetch logging changes

tsightler:
Hi @dgreif,

For you consideration, here are a few tweaks to the default logging behavior of ring-client-api around fetch requests.  This PR consist of the following changes:

1) Single fetch failures are no longer logged, only retries, hopefully reducing log spam from spurious API failures.
2) Any "fetch is not defined" errors also log information about unsupported NodeJS version
3) Fetch errors now include the retry number as well as the cause message and code (if they exist) even when debug is not enabled.  The new error would look something like this:

`Retry #4 failed to reach Ring server at https://api.ring.com/clients_api/ring_devices.  Error: fetch failed, Cause: Connect Timeout Error (attempted addresses: 15.197.187.26:443, 3.33.190.236:443), Code: UND_ERR_CONNECT_TIMEOUT.  Trying again in 5 seconds...`

This provides a bit more information which I'm hopeful might be useful to track down some of these ongoing issues with failing to connect without requiring debug which clutters the log with full stack traces.

Open to any feedback here.

### DIFF
--- a/packages/ring-client-api/rest-client.ts
+++ b/packages/ring-client-api/rest-client.ts
@@ -93,6 +93,12 @@ async function requestWithRetry<T>(
   requestOptions: RequestOptions & { url: string; allowNoResponse?: boolean },
   retryCount = 0,
 ): Promise<T & ExtendedResponse> {
+  if (typeof fetch !== 'function') {
+    throw new Error(
+      `Your current NodeJS version (${process.version}) is too old to support this plugin.  Please upgrade to the latest LTS version of NodeJS.`,
+    )
+  }
+
   try {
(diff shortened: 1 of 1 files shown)


tsightler:
Included one more change, use a custom agent to disable keepalive and increase the connection level timeout to 20 seconds as well (currently, code only increases the request timeout, connect timeout is shorter).  My hope is this might address some of the "failed to reach server" reports as there are some similar cases in Node and Undici Github projects.  Mostly this is just a test.

dgreif (approved):
Love it, thanks for getting all these tweaks in!

PR #1532, opened 2024-11-21
Tune fetch for improved connection management

tsightler:
Hi @dgreif,

This PR is an attempt to address the native fetch/undici issues that were causing http requests to stop.  I can provide a hugely detailed analysis of the failure, but, suffice it to say, I believe it has to do with how undici uses keepalive and pool connections.  There are quite a number of similar issues on the undici project page.

To work around this, the PR imports Agent from the undici package and creates a custom dispatcher with fewer max TCP connections and much longer keepalive.  This makes native NodeJS fetch behave more like browser fetch implementations as well as other http client libs like got (the actual values used are the same as Firefox).  This actually has a mostly positive impact for all users as it dramatically reduces the number of DNS lookups and TCP handshakes, since the HTTP connections are now long lived.

So far, I've had 4 users that were experiencing the issue test this patch, and success has been 100%.  I've tried to keep the patch as minimal as possible, with only the absolute minimum changes required.

In theory, it should even be possible to do this without pulling in undici dependency as it is possible to modify the global dispatcher, but I had some challenges with that and it felt very hackish.  However, if you believe this is better, I can pursue it.

Also, with a custom dispatcher it is possible to enable HTTP/2 via the `allowH2: true` option.  I did test this, however, H2 support in Undici has some pretty ciritcal bugs that regularly caused the NodeJS process to crash (issues are open, hopefully fixed in the next Undici release).  I'll monitor this for the future, but it doesn't seem stable for the time being.

### DIFF
--- a/.changeset/purple-teachers-applaud.md
+++ b/.changeset/purple-teachers-applaud.md
@@ -1,0 +1,6 @@
+---
+'ring-client-api': patch
+'homebridge-ring': patch
+---
+
+Increase keepalive and adjust connection pooling, which will hopefully result in more stable fetch requests
--- a/package-lock.json
+++ b/package-lock.json
@@ -1459,9 +1459,9 @@
       }
     },
     "node_modules/@eslint/plugin-kit": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/@eslint/plugin-kit/-/plugin-kit-0.2.0.tgz",
-      "integrity": "sha512-vH9PiIMMwvhCx31Af3HiGzsVNULDbyVkHXwlemn/B0TFj/00ho3y55efXrUZTfQipxoHC5u4xq6zblww1zm1Ig==",
+      "version": "0.2.3",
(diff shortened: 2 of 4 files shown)


tsightler:
BTW, this is the main blocker for HTTP/2 for now is https://github.com/nodejs/undici/issues/3753.  Once it is fixed we can evaluate it's stability again, but we'd need a release of NodeJS that has fixed undici, or we'd have to move to using fetch from undici instead, which is also very easy, just a few minor tweaks to this are needed, if that is preferred.

For now I stayed with native fetch with custom dispatcher instead because my understanding is that there is a plan to expose dispatcher in future Node versions so eventually we can drop undici dependency.  Happy to do it either way.

dgreif (approved):
Thanks for your all your research and testing on this! I'm very happy we landed on a solution which allows us to use _longer_ keep-alives, rather than terminating every connection. It's unfortunate we have to directly depend on undici, but that's really a quirk of how node is handling this. I'll get this merged and cut a beta release.

I'm also good with keeping this approach, rather than tweaking undici globally. That might impact other things (e.g. plugins) running in the same node instance, which we probably don't want to do 😄

PR #1557, opened 2024-12-29
Add support for Battery Video Doorbell Pro

hkoci:
I recently purchased the Pro version of the Battery Video Doorbell - I am trying to get the actual device model to show up on the addons I use, Scrypted, Home Assistant (etc.) it makes the device look better rather it saying unknown model in Homekit 😋 

I presume just the v3 visual identifier needs to be added in the types - functionality is working 👍🏻


hkoci:
There is also a cocoa_doorbell_v2, I do not own this device - I think this is a Battery Doorbell Plus

dgreif (approved):
Thanks for getting these added!

PR #1589, opened 2025-02-12
Doorbell Notifications Control Ring Intercom

nd3442:
Added ability to turn off Homekit Doorbell Notifications with a switch for Ring Intercom.

Issue 1586


dgreif:
@nd3442 after reviewing the related issue, it sounds like you simply need to remove your intercom from HomeKit since you don't want the related notifications, and that's about all you get for intercoms today. You can achieve this with the existing `hideDeviceIds` option. See the [readme](https://github.com/dgreif/ring/tree/main/packages/homebridge-ring#optional-parameters) for more details

PR #1650, opened 2025-09-01
feat(homebridge-ring): add allowDisarm option to block disarm from HomeKit

salmanesm:
Summary
This PR adds a new config flag, allowDisarm (default: true), to the Homebridge Ring plugin. When set to false, HomeKit Disarm requests are quietly blocked, while Arm Home/Away/Night continue to work as usual.

Why
Throwing an error from the .onSet handler surfaced as an error in the Home app/Siri. That confused users even though the block was intentional.
This change converts the behavior to a no-op + snap-back: disarm requests are logged and ignored, and the Home app UI immediately reverts to the prior state without showing a red error banner.

What changed

Config option: allowDisarm?: boolean (default = true, backward compatible).

security-panel.ts:

If state === DISARM and allowDisarm === false, log a warning and update characteristics back to their prior values.

Return early (no Ring API call, no error thrown).

Docs: README updated with config snippet.

Behavior

allowDisarm: true: unchanged (HomeKit can arm and disarm).

allowDisarm: false:

Disarm requests are ignored.

UI/Siri behave as if nothing changed (no error toast).

Ring stays armed.

### DIFF
--- a/packages/homebridge-ring/README.md
+++ b/packages/homebridge-ring/README.md
@@ -59,6 +59,7 @@ Only include an optional parameter if you actually need it. Default behavior wit
 {
   "alarmOnEntryDelay": true,
   "beamDurationSeconds": 60,
+  "allowDisarm": false,
   "hideDeviceIds": [
     "477e4800-fcde-4493-969b-d1a06f683102",
     "5aaed7a7-06df-4f18-b3af-291c89854d60"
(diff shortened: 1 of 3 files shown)


salmanesm:
Hi @dgreif — this adds an optional `allowDisarm` flag (default `true`). 
When `false`, HomeKit “Disarm” is quietly ignored (no error toast), while arming modes still work. 
Docs updated with a config example. Happy to adjust naming, default, or placement if you prefer!

salmanesm:
Config used for validation: "allowDisarm": false
Expected behavior:
• Arm Home/Away/Night succeed
• Disarm is ignored (no error toast)
• Log: [Ring Alarm] Disarm attempt from HomeKit blocked (allowDisarm=false)
Implementation resets both TargetState and CurrentState so the Home app stays in sync.

dgreif:
Thanks for the PR @salmanesm! I see the error banner as a "feature", not a bug. What are the scenarios where you expect a disarm to fail and are ok with it?

PR #1653, opened 2025-09-01
feat(homebridge-ring): expose Ring Alarm panel volume as a HomeKit slider (opt-in)

salmanesm:
## Summary
Expose the Ring Alarm **Base Station / Panel** volume to HomeKit as a visible slider. This is **opt-in** and coexists with the existing Speaker/Volume characteristic.

## Why
- Quick volume adjustments from the Home app (no need to open Ring).
- Enable scenes/automations (e.g., “Night” → quieter chimes).
- Keep defaults unchanged for users who don’t want the extra tile.

## Implementation
- **New accessory:** `VolumeAccessory`
  - Uses **Lightbulb** service so the slider is prominent.
  - Maps **Brightness (0–100)** ↔ **volume (0.0–1.0)**.
  - `On` = volume > 0; turning `Off` sets volume to 0.
  - Subscribes to `device.onData` to keep HomeKit in sync.
  - Gracefully handles devices without `setVolume()`.
- **Platform wiring:** in `ring-platform.ts`, when `exposeAlarmVolume` is true, attach `VolumeAccessory` for devices that have `data.volume` and `setVolume()`.
- **Config:** adds `exposeAlarmVolume?: boolean` (default `false`), and reserves `volumeService?: 'lightbulb' | 'speaker'` for future use.
- **NodeNext-friendly imports:** use explicit `.ts` extensions.

## Behavior
- **Default:** no changes unless `exposeAlarmVolume` is enabled.
- When enabled, a new accessory like “\<Panel Name\> Volume” appears with a slider.
- Slider changes affect the device immediately; external volume changes reflect back in HomeKit.

## Config example
```json
{
  "platform": "Ring",
  "refreshToken": "<your-token>",
  "exposeAlarmVolume": true
}

### DIFF
--- a/packages/homebridge-ring/config.ts
+++ b/packages/homebridge-ring/config.ts
@@ -23,6 +23,8 @@ export interface RingPlatformConfig extends RingApiOptions {
   onlyDeviceTypes?: string[]
   showPanicButtons?: boolean
   disableLogs?: boolean
+  exposeAlarmVolume?: boolean // Expose Base Station / Keypad volume as a HomeKit slider
+  volumeService?: 'lightbulb' | 'speaker' // (reserved for future) choose service type
 }
 
 export function updateHomebridgeConfig(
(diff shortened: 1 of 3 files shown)


dgreif:
@salmanesm thanks for the PR. Could you walk me through your use case for adjusting the alarm volume? I generally try to avoid adding new options as they make maintenance and future changes harder. I'd also be cautious of adding this as a lightbulb as it makes certain siri commands more unpredictable (e.g. "turn off all the lights" will now slide your volume to zero 😬).

PR #1749, opened 2026-05-25
Update ring-client-api to use v3 location/devices REST endpoints

tsightler:
This is the first in a series of PRs to bring ring-client-api on par with modern Ring API specs.  While I have tested these changes with ring-mqtt, these are changes in core functions that have remained largely unchaged for many years so this PR has more risk than many.

This PR updates the Ring client library to use the latest REST API endpoints for location and device discovery.  I've tried to keep the changes to the minimum and not break anything for API consumers.  The new locations endpoint provides roughly the same information and format of the previous, however, the new devices endpoint is quite different.

I've tested this with my account which has the following devices:

Doorbell Pro
Floodlight (1st gen)
Stickup Cams (various)
Floodlight Pro
Chime
Alarm (1st gen) with many devices

The most likely risks to the change are around shared locations as there are some differences in how discovery of location devices works, but I don't have a shared account to test.

**Important Note:**  This code was written 99% by Claude Code.  I have manually reviewed the code, applying a few minor tweaks, mostly for style, and also performed manual functional testing.

### DIFF
--- a/packages/ring-client-api/api.ts
+++ b/packages/ring-client-api/api.ts
@@ -1,5 +1,10 @@
 import type { RefreshTokenAuth, SessionOptions } from './rest-client.ts'
-import { clientApi, deviceApi, RingRestClient } from './rest-client.ts'
+import {
+  clientApi,
+  deviceInfoApi,
+  locationInfoApi,
+  RingRestClient,
+} from './rest-client.ts'
 import { Location } from './location.ts'
 import type {
   BaseStation,
@@ -14,7 +19,12 @@ import type {
   UnknownDevice,
   UserLocation,
 } from './ring-types.ts'
-import { PushNotificationAction, RingDeviceType } from './ring-types.ts'
+import {
+  ChimeModel,
+  PushNotificationAction,
+  RingCameraKind,
+  RingDeviceType,
+} from './ring-types.ts'
 import type { AnyCameraData } from './ring-camera.ts'
 import { RingCamera } from './ring-camera.ts'
 import { RingChime } from './ring-chime.ts'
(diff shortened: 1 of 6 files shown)


packages/ring-client-api/api.ts
```diff
@@ -38,6 +48,17 @@ import { PushReceiver } from '@eneris/push-receiver'
...
+const doorbellKinds: Set<string> = new Set(
+  Object.values(RingCameraKind).filter(
+    (k) =>
+      k.startsWith('doorbot') ||
+      k.startsWith('doorbell') ||
+      k.startsWith('lpd_') ||
+      k.startsWith('jbox_') ||
+      k.startsWith('cocoa_doorbell'),
+  ),
+)
```
tbaron, line 60: This could be cleaner to encapsulate in an exported predicate in `ring-types.ts`, similar to `isWebSocketSupportedAsset` that's already implemented there. If placed in proximity to the `RingCameraKind` type, it would also improve maintenance discoverability. E.g.,

```ts
export type RingCameraKind = ...

export function isDoorbellKind({ kind }: { kind: RingCameraKind }) {
    return kind && (kind.startsWith( ... ) || ... );
}
```

Or more tersely with a bound regex `.test()`:

```ts
export const isDoorbellKind =
RegExp.prototype.test.bind(/^(?:(cocoa_)?doorb(ell|ot)|lpd_|jbox_)/) as
(kind: RingCameraKind) => boolean
```

packages/ring-client-api/rest-client.ts
tbaron, line 61: Should the remaining `/devices/v1/` usages be deprecated in favor of `/device_info/v3/`?

packages/ring-client-api/ring-types.ts
```diff
@@ -429,13 +429,21 @@ export interface LocationAddress {
...
   owner_id: number
   updated_at: string
   user_verified: boolean
+  is_owner?: boolean
+  operation_set?: string
+  entitlements?: {
+    business_command_center?: boolean
+    business_role_management?: boolean
+    vm_virtual_security_guard?: boolean
+  }
```
tbaron, line 445: Curious - why are these required for the broader change? Could the type spec be generalized to allow flexible parsing of future changes (e.g., key-value map)?

tbaron (commented):
Thanks for making these changes! Did you find any documentation for the v3 APIs, or were these changes reverse engineered? I didn't see them in the [official docs](https://developer.amazon.com/docs/ring/api-documentation.html#endpoint-reference).

Issue #1681, opened 2025-11-19
Heavy log traffic


sabresoftware2021:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe The Bug

Not sure if this is an issue. The plugin seems to be working OK, but the HomeBridge log file has a lot of entries as shown below:

16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[16/11/2025, 19:18:21] [Ring] Ignoring push notification received in first two seconds after starting up
[...]

PR #1799, opened 2026-07-17
fix(ring-client-api): summarize ignored startup push notifications

### LINKED ISSUE #1800: Summarize ignored startup push notifications instead of logging each one
### Description

On Homebridge restart, Ring delivers a burst of queued push notifications during the intentional 2-second startup ignore window in `ring-client-api`. Each ignored notification is logged at info level, producing dozens of identical lines:

```
[homebridge-ring] Ignoring push notification received in first two seconds after starting up
```

This matches closed #1681. The ignore behavior is correct; the logging is noisy for accounts with many devices/events.

### Proposed solution

Keep the 2s ignore window. Count ignored notifications and emit a single summary after the window, e.g. `Ignored 63 push notification(s) received in the first 2 seconds after starting up`.

PR: #1799

### Alternatives

- Log at debug only (harder to notice when troubleshooting)
- Persist push message IDs (larger change; what the ignore window is avoiding)

### Device data

N/A — logging-only change; no new device data.

tbaur:
## Summary
- During the intentional 2s startup push-ignore window, Ring often delivers dozens of queued notifications.
- Each ignored notification was logged at info level, flooding Homebridge logs on every restart.
- Count ignores and emit a single summary line after the window (e.g. `Ignored 63 push notification(s)...`).

Fixes #1800

Related: #1681 (closed as not planned when framed as a bug; this keeps the ignore behavior and only reduces log noise).

## Test plan
- [x] `npm run lint -w ring-client-api`
- [x] `npm run test -w ring-client-api`
- [ ] Restart homebridge-ring and confirm one summary line instead of many identical ignore messages

### DIFF
--- a/.changeset/summarize-startup-push-ignores.md
+++ b/.changeset/summarize-startup-push-ignores.md
@@ -1,0 +1,5 @@
+---
+"ring-client-api": patch
+---
+
+Summarize ignored startup push notifications into a single log line instead of one line per notification.
--- a/packages/ring-client-api/api.ts
+++ b/packages/ring-client-api/api.ts
@@ -27,6 +27,7 @@ import {
 } from 'rxjs/operators'
 import {
   clearTimeouts,
+  delay,
   enableDebug,
   logDebug,
   logError,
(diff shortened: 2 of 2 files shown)


tbaur:
Opened tracking issue for this change.

PR #1810, opened 2026-08-05
fix(ring-client-api): FCM reliability - tighter heartbeat, no dropped reconnect pushes

sappkevin:
Hey! This is the companion fix to Eneris/push-receiver#40. Same root problem, different layer.

**1. The heartbeat was too relaxed for home networks**

Default is `heartbeatIntervalMs: 300_000` (5 minutes), and paired with the underlying socket's default keepalive (7200 seconds on Linux, see the linked PR), a dead connection on a residential NAT could sit undetected for 2+ hours. I bumped it to 60 seconds so a stale connection gets noticed and reconnected within a couple of minutes instead.

**2. The "ignore first 2 seconds of pushes" guard was eating real events**

This one bit me directly. `registerPushReceiver` only runs once per `RingApi` instance (at startup), so this guard's 2-second window only ever applied right after a restart, not on every internal FCM reconnect within a session. But that was still enough to cause real damage: one restart dropped 23 queued overnight motion events, several confirmed real (raccoon on camera, zero alert). push-receiver already dedupes by `persistentId` within a session, so the window wasn't preventing duplicates, it was just discarding whatever backlog happened to be waiting right as the listener came online.

Removed the guard entirely. Only touches `api.ts`, no public API changes.

**Testing**

Ran this against a real Ring account for several days before and after. Before: reproducible multi-hour silences after a NAT idle timeout. After: self-recovers in a minute or two. Also directly tested the reconnect-drop fix by forcing a reconnect during active motion and confirming the event now comes through instead of getting swallowed.

Added `test/api.spec.ts` covering the heartbeat value and both notification-delivery behaviors (startup delivery, no new drop window on disconnect/reconnect). 21/21 tests and lint pass locally.

---

Thanks to [@danieloleary](https://github.com/danieloleary) for the review: he wrote the test file above, caught an unused `logInfo` import the guard removal left behind, and correctly called out that my original wording overstated the old guard as dropping events "on every reconnect" when it's really only at startup. Both folded in, with credit in the commits.

### DIFF
--- a/.changeset/fcm-heartbeat-and-drop-guard.md
+++ b/.changeset/fcm-heartbeat-and-drop-guard.md
@@ -1,0 +1,5 @@
+---
+"ring-client-api": patch
+---
+
+Tighten FCM push reliability: lower the PushReceiver heartbeat from the 5 minute default to 60 seconds, and stop dropping notifications received in the first 2 seconds after connecting. Together these fix silent, multi-hour motion-push delivery stalls observed on residential NAT connections, where a dead socket could otherwise go undetected for 2+ hours and queued reconnect events were being discarded instead of delivered.
--- a/packages/ring-client-api/api.ts
+++ b/packages/ring-client-api/api.ts
@@ -25,13 +25,7 @@ import {
   switchMap,
   throttleTime,
 } from 'rxjs/operators'
-import {
-  clearTimeouts,
-  enableDebug,
-  logDebug,
-  logError,
-  logInfo,
-} from './util.ts'
+import { clearTimeouts, enableDebug, logDebug, logError } from './util.ts'
 import { setFfmpegPath } from './ffmpeg.ts'
 import { Subscribed } from './subscribed.ts'
 import { PushReceiver } from '@eneris/push-receiver'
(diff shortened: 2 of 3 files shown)


danieloleary:
Hey Kevin — this is a great find. The raccoon story got me 😂

I pulled your branch and wrote a few tests around the heartbeat and startup notification behavior. I also removed a leftover `logInfo` import that was tripping lint.

If helpful, you can grab them here:
- [tests](https://github.com/danieloleary/ring/commit/f6ee3ed)
- [lint cleanup](https://github.com/danieloleary/ring/commit/0073a9a)

Everything passes here: 21 tests, lint, and build.

One tiny wording thing I noticed while I was in there: the two-second guard seems to run only at startup, not on every reconnect. And the existing heartbeat appears to retry after about ten minutes, so “faster recovery” may be the safest claim. Neither changes the value of the fix.

No separate PR from me. Just wanted to help you get this over the line. Nice work.

sappkevin:
Thanks Dan, this is great. Pulled in both commits (with co-author credit) and fixed the "every reconnect" wording per your catch, since you're right that the guard only ever ran once at startup. Tests are in, 21/21 passing, lint clean. Appreciate you taking the time.

PR #1193, opened 2023-04-04
fix: 406 error

rvmoldova:
Hi,

Adding `User-Agent` header fixes issue mentioned in #1192 

```
TypeError: Cannot use 'in' operator to search for 'error' in <html>
<head><title>406 Not Acceptable</title></head>
<body bgcolor="white">
<center><h1>406 Not Acceptable</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

PR #1818, opened 2026-08-24
fix(ring-client-api): handle non-JSON oauth error responses

### LINKED ISSUE #1817: getAuth() throws TypeError when the oauth endpoint returns a non-JSON error body
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe The Bug

`getAuth()` assumes the error response body is JSON:

```ts
// rest-client.ts:367
responseData: Auth2faResponse = response.body || {},
responseError =
  'error' in responseData && typeof responseData.error === 'string'
```

`responseToError` always attempts JSON.parse and keeps the raw text on failure, so response.body can be a string. 'error' in <string> then throws, and the error handler dies instead of reporting the failure:

```
TypeError: Cannot use 'in' operator to search for 'error' in <html>...
```

Any non-JSON error body triggers it. A proxy's 406 Not Acceptable HTML page is the usual one. Same crash is quoted in the body of the PR #1193, so it isn't new.

Three reads of responseData are affected: 
1. 'error' (:369)
2. 'tsv_state' (:385)
3. 'error_description' (:409)

### To Reproduce

Point the oauth endpoint at anything returning a non-JSON body with an error status, then call `getAuth()`. 

```ts
http.post('https://oauth.ring.com/oauth/token', () =>
  HttpResponse.text('<html><head><title>406 Not Acceptable</title></head></html>', { status: 406 }),
)
```

### Expected behavior

The error handler should report the failure rather than throw. Something like 

```
Failed to fetch oauth token from Ring. ... (error: HTTP 406)
```

Instead of a TypeError.

### OS

Ubuntu 24.04.4 LTS (WSL2

### Node.js Version

v22.22.1

### NPM Version

v10.9.4

### ring-client-api

v14.3.0

### Operating System

Ubuntu in WSL2

SirLouen:
Closes #1817 

A non-JSON body from the OAuth endpoint (`406 Not Acceptable` HTML page) makes `getAuth()` throw `TypeError: Cannot use 'in' operator to search for 'error'` from its own error handler. Same crash reported in #1193.

The body is now only treated as JSON when it is an object, and `responseError` falls back to `HTTP <status>`. 

That fallback also applies to JSON error bodies with no string `error` field, which used to end in `(error: )`.

The new test uses `server.use`, so `afterEach` now calls `server.resetHandlers()`.

**Relevant to #1704**: it relocates this block ahead of the refresh-token branch, which would expose refresh-token clients to the same crash.

### DIFF
--- a/.changeset/fix-non-json-oauth-error.md
+++ b/.changeset/fix-non-json-oauth-error.md
@@ -1,0 +1,5 @@
+---
+"ring-client-api": patch
+---
+
+Fix a `TypeError` in `getAuth()` when the oauth endpoint returns a non-JSON error body (for example a `406 Not Acceptable` HTML page); the failure is now reported as `(error: HTTP 406)`.
--- a/packages/ring-client-api/rest-client.ts
+++ b/packages/ring-client-api/rest-client.ts
@@ -364,10 +364,15 @@ export class RingRestClient {
       }
 
       const response = requestError.response || {},
-        responseData: Auth2faResponse = response.body || {},
+        responseData: Auth2faResponse =
+          typeof response.body === 'object' && response.body !== null
+            ? response.body
+            : {},
         responseError =
           'error' in responseData && typeof responseData.error === 'string'
             ? responseData.error
+            : response.status
+            ? `HTTP ${response.status}`
             : ''
 
       if (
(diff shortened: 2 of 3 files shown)


dgreif (approved):
Changes look good, thank you!

PR #1821, opened 2026-09-01
feat(intercom): two-way audio for the Ring Intercom

### LINKED ISSUE #1820: Ring Intercom: two-way audio is available but never opened (stealth mode)
## Summary

The Ring Intercom's audio channel is available through the existing streaming stack, but it is never opened, so in HomeKit you can unlock the door and receive the ding while being unable to hear who is there or speak to them.

I have this working on my own intercom and would like to know whether you would want it upstream before I take up your review time. A PR is ready either way, and I am equally happy to keep it in my fork if this is not a direction you want the plugin to go.

## What I found

The intercom is a doorbot like the cameras, the streaming ticket (`clap/ticket/request/signalsocket`) is generic and carries no device id, and negotiating WebRTC with the intercom's `doorbot_id` makes the server answer:

```
m=audio 9 UDP/TLS/RTP/SAVPF 96
a=sendrecv
```

Audio in both directions, no video track.

**The blocker is stealth mode.** An intercom starts with its microphone closed and only opens it on receiving `camera_options { stealth_mode: false }`. `activateCameraSpeaker()` sends exactly that, but it is currently only called once HomeKit starts *returning* audio — that is, when the user presses the microphone to talk. Until that moment the device transmits digital silence, whether or not there is an active call.

Measured per-block RMS, changing nothing else:

```
without stealth_mode:false → -28 -74 -80 -85 -92 -92   (digital silence)
with    stealth_mode:false → -42 -51 -53 -58 -64 -69   (continuous signal)
```

(The -28 is a connection click, not audio — which is why these are per-block rather than a max.)

## The other half: video

HomeKit will not accept a camera without a video track, and the intercom has no camera. Two things are needed:

1. A video track has to be synthesised. A small bundled still encoded to H.264 by ffmpeg works, at ~1.5% of one core on a Pi 5.
2. Ring's video **must not** be forwarded. For an intercom that channel does not carry valid H.264, and HomeKit tears down the entire session on receiving it — taking the audio with it. This is why simply reusing the camera path does not work.

## Question

Would you want this in the plugin? If so I will open the PR against `main`. It is opt-in and off by default, and the existing intercom accessory is untouched — the audio is published as a separate unbridged accessory, since converting the current one would force users to re-pair and lose their automations.

Working branch, if you want to look before deciding: https://github.com/mruwzum/ring/tree/intercom-two-way-audio

mruwzum:
Closes #1820 — please read that issue first; it explains the finding. Opening the PR alongside it so the code is there if you want it, but I am not expecting review time until you have said whether you want the feature at all. Happy to close this and keep it in my fork.

## What this does

Adds two-way audio for the Ring Intercom, plus three smaller things the device already exposed and the plugin did not use. **Everything is off by default** and no existing accessory changes.

| Option | Default | |
|---|---|---|
| `enableIntercomAudio` | `false` | Camera accessory with a still image and live two-way audio |
| `showOfflineSensor` | `false` | Contact sensor from `alerts.connection` |
| `showDoNotDisturbSwitch` | `false` | Unsubscribes from dings on Ring's servers |
| `logIntercomDings` | `false` | Unthrottled JSONL log of dings and unlocks |
| `intercomMicGainDb` / `intercomSpeakerGainDb` | `12` / `10` | Gain, with a limiter on both |

Also: `hideDoorbellSwitch` is now honored for the intercom. It was documented and ignored there, so anyone who set it still got the switch.

## How the audio works

Covered in #1820. The short version: the intercom starts in stealth mode and transmits silence until it receives `camera_options { stealth_mode: false }`, which today is only sent when the user presses the microphone. This sends it when the stream opens.

For video, a bundled 320x240 still is encoded to H.264 by ffmpeg. Ring's video is **not** forwarded — for an intercom it is not valid H.264 and HomeKit drops the whole session, audio included.

## Notes for review

**`intercom-camera-source.ts` is a copy of `camera-source.ts`, not a subclass.** The class that needs changing, `StreamingSessionWrapper`, is not exported, and I did not want to touch a file that real cameras depend on. If you would rather I export it and subclass, or refactor the shared parts out, say so and I will redo it that way — I took the conservative option because breaking camera streaming to add an intercom feature would be a bad trade.

**The audio accessory is unbridged and separate.** HomeKit requires cameras to be unbridged, and converting the existing intercom accessory would force users to re-pair and lose their automations.

**Do Not Disturb is deliberately not persisted.** After a restart it returns to "receiving dings". A DND that forgets itself is a much better failure than one that silently costs someone their doorbell.

## Tests

13 cases in `test/intercom.spec.ts`, using vitest to match `ring-client-api`. This adds a `test` script to the `homebridge-ring` package, which had none.

Five are regressions of failures I actually hit, each of which left me with no audio and **no error message anywhere**:

- **10 fps is a floor.** I dropped it to 5 to save CPU; HomeKit began cutting sessions after a few seconds and audio fell from ~25 KiB to 2 KiB. The optimization that works is frame *size*.
- Stealth mode not being disabled at stream start.
- Audio timestamps: without `genpts`/`aresample`, ffmpeg reports "Queue input is backward in time" and HomeKit discards audio that arrives fine.
- Forwarding Ring's video.
- An unset ffmpeg path, which surfaces only as "not responding".

The build now runs `rm -rf ./lib/test`, matching `ring-client-api`, so tests do not ship in the published package.

## Verification

Running on a Raspberry Pi 5 against a real Ring Intercom: audio verified in both directions from the Home app, including from the street. 3.7% of one core with a live stream, no temperature rise over idle.

`turbo run build test lint` passes across the monorepo.

## Docs

`packages/homebridge-ring/INTERCOM.md`, including the measurements and the two things that will silently break it for the next person.


mruwzum:
Adding the device/account detail the [Common Feature Requests](https://github.com/dgreif/ring/wiki/Common-Feature-Requests) page asks for, since I had left it out.

**Tested against:** a single Ring Intercom reporting `intercom_handset_audio`, on a personal (non-shared) Ring account with one location and no Ring Protect subscription. Homebridge 2.4.0 / HAP 2.2.2 on a Raspberry Pi 5 (Debian 13, Node 24, ffmpeg 7.1.5), clients on iOS.

**Not tested, and the PR does not claim to support it:** `intercom_handset_video`, shared/guest accounts, or multiple intercoms on one location. The code paths are generic over `location.intercoms`, but I only have the one device, so I would not want that read as verified.

I take the point about scope. Two things I can offer:

- If the size is the blocker, the audio is separable from the rest. The connectivity sensor, Do Not Disturb and ding log are ~120 lines against `intercom.ts` and carry none of the streaming risk — happy to split those into their own PR and drop the audio, if that split is more reviewable.
- If the copied `camera-source.ts` is the blocker, I would rather export `StreamingSessionWrapper` and subclass it. I avoided that only because I did not want to touch a file real cameras depend on without a maintainer saying it was fine.

Two fixes pushed since opening, both found by running it rather than reading it:

- `hasSnapshotWithinLifetime` was a method where `RingCamera` has a getter. `camera-source` reads it as a property, so the negation was always false and the snapshot never loaded — the only symptom was `No snapshot cached` on every request. Regression test added.
- The Mute `onGet` override was fighting the handler `configureController` already installs, producing a HAP warning per request. Now only missing services are added.

No pressure either way — a "no" here is a perfectly good outcome and I will keep it in my fork.

mruwzum:
Closing this myself — the triage answer on #1820 was clear that intercom audio is out of scope, and the wiki asks for small PRs where this is 1400 lines. No point taking up review time on something already answered.

Leaving it here for anyone who lands on the same problem: the working code is at https://github.com/mruwzum/ring/tree/intercom-two-way-audio, and the finding that matters is in #1820 — an intercom transmits digital silence until it receives `camera_options { stealth_mode: false }`, which is currently only sent when the user presses the microphone to talk.

Thanks for the plugin. It has been running my house for a long time.

